PR #3, opened 2017-06-30
Remove no longer needed week-glance-session-attrs component

jrjohnson:
This data is now available in the event object directly so it can be
loaded from there without accessing the session directly.


saschaben:
attire / equipment icons are a go.

It appears we are no longer identifying supplemental/required. we need to address this.

jrjohnson:
I can add that to this PR, what icon would you like to use for that?

saschaben:
I think this one just gets a label: `Required`, in bold small caps. I can't think of a reasonable icon for this stuff.

saschaben:
come to think of it. ....since the "on" value of this attribute is "supplemental", THAT should be what we are displaying. Not required (which might get confused with our attendance required attribute. So maybe when on, a `supplemental` flag...?

jrjohnson:
Supplemental is a long word and we have very limited space here where we have all of the icons.  What if we do a star for not supplemental and an empty star for supplemental. So either way there would be a star, but for supplemental it would be open.
http://fontawesome.io/icon/star-o/

Then we can tooltip them with "required" / "supplemental" like the other icons.

saschaben:
I think we need some only from the users on this -- let's leave it be for now, and make a separate ticket until we are sure of the right iconography.

saschaben:
so...let's close this one out. 

👍  APPROVED.

jrjohnson:
Deploy preview ready!

Built with commit 0ed35953859bd5cbd2d7b76acbdcf5a81159bc49

https://deploy-preview-3--ilios-common.netlify.com

PR #10, opened 2017-07-26
Remove searchString CP from session

jrjohnson:
This isn't useful anymore since status is no longer provided by
PublisableModel. We need to construct the search string in the calling
component instead where we will have access to i18n.


jrjohnson:
Deploy preview ready!

Built with commit ebb5657cc56eb08fbc7dafaac1769aa6b2a82bc6

https://deploy-preview-10--ilios-common.netlify.com

PR #28, opened 2017-08-15
Display weekly calendar with no interaction

jrjohnson:
When we want to display the calendar, but not make anything live we can
use these options.

Refs ilios/frontend#2061


jrjohnson:
Deploy preview ready!

Built with commit 9ccaff00f77856368239e8a518280b1f4aca9b2f

https://deploy-preview-28--ilios-common.netlify.com

PR #30, opened 2017-08-15
Pass selectable status down to multiday events

jrjohnson:
This status is used to determine if click actions are available and
should be the same for a single event on a day and a multi day event.


jrjohnson:
Deploy preview ready!

Built with commit ffc67ed24fb8e8d27c3a71ee8ff4a351dfc7417b

https://deploy-preview-30--ilios-common.netlify.com

PR #38, opened 2017-08-21
Switch to Yarn for dependency management

### LINKED ISSUE #19: add yarn lock file
notice that we version control the yarn lock file in frontend, but not here. 

i'm proposing we add it to this project as well.

@jrjohnson  thoughts/objections?

jrjohnson:
Since yarn appears to work now for libsass we should move to it so all of our packages are on the same system.

Fixes #19


jrjohnson:
Deploy preview ready!

Built with commit 0e3a22a6868dfcc4db35c158e58f927a92f27971

https://deploy-preview-38--ilios-common.netlify.com

PR #41, opened 2017-08-27
Remove limits that were not real

### LINKED ISSUE #40: Remove API Limits
Now that ilios/ilios#1910 has landed and the API no longer has default limits we should strip all of our limit code our of the frontend where it was input simply to ensure we got all data {limit: 1000}, {limit: 100000}` etc...

jrjohnson:
These limits were being used to ensure that we got all data when the API
had a default limit of 20 items. They are no longer necessary since the
API now returns all data starting in v1.23.

Fixes #40


jrjohnson:
Deploy preview ready!

Built with commit a00f28cde51ec76cac0445d37a22db567867ff15

https://deploy-preview-41--ilios-common.netlify.com

PR #88, opened 2017-10-17
Correct the proportions of the pulse-loader SVG

jrjohnson:
There was a lot of unused space in the box for this SVG so it was
difficult to use it in smaller areas. I constrained the viewBox to the proportions needed by the animation.  I also removed some CSS we were no longer using and text-align centered the loader which looks nicer.


jrjohnson:
Deploy preview ready!

Built with commit 98cbbf6e869d6908790a8f2bd189e17f0254bc56

https://deploy-preview-88--ilios-common.netlify.com

PR #89, opened 2017-10-17
use comp properties correctly.

### LINKED ISSUE #90: don't rely on non-public `content` attribute when doing existence checks
```
$ grep -r '\.content' addon
addon/models/term.js:  isTopLevel: empty('parent.content'),
addon/models/term.js:  hasChildren: notEmpty('children.content'),
addon/models/session.js:  isIndependentLearning: notEmpty('ilmSession.content'),
addon/models/learner-group.js:    if(this.get('parent.content')){
```

stopfstedt:
fixes #90

### DIFF
--- a/addon/models/learner-group.js
+++ b/addon/models/learner-group.js
@@ -172,7 +172,7 @@ export default DS.Model.extend({
 
     return title;
   }),
-  allParentTitles: computed('parent.{title,allParentTitles}', function(){
+  allParentTitles: computed('isTopLevelGroup', 'parent.{title,allParentTitles}', function(){
     let titles = [];
     if(!this.get('isTopLevelGroup')){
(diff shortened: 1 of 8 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 6caafc6795e655e5e00fddae273b36c3fc2727fd

https://deploy-preview-89--ilios-common.netlify.com

addon/models/learner-group.js
```diff
@@ -174,7 +174,7 @@ export default DS.Model.extend({
   }),
   allParentTitles: computed('parent.{title,allParentTitles}', function(){
```
jrjohnson: Should depend on the `isTopLevelGroup` property.

stopfstedt: good catch. fixed!

Issue #79, opened 2017-10-03
replace RSVP.Promise usage with async/await


stopfstedt:
```
$ grep -r Promise addon | grep RSVP
addon/mixins/categorizable-model.js:const { Promise, all } = RSVP;
addon/initializers/replace-promise.js:  window.Promise = RSVP.Promise;
addon/models/competency.js:const { Promise, all } = RSVP;
addon/models/program.js:const { all, Promise} = RSVP;
addon/models/course.js:const { all, map, Promise } = RSVP;
addon/models/curriculum-inventory-sequence-block.js:const { Promise } = RSVP;
addon/models/term.js:const { Promise } = RSVP;
addon/models/session.js:const { all, Promise } = RSVP;
addon/models/offering.js:const { Promise, all } = RSVP;
addon/models/curriculum-inventory-report.js:const { Promise, all } = RSVP;
addon/models/school.js:const { all, Promise } = RSVP;
addon/models/learner-group.js:const { Promise, map, all } = RSVP;
addon/models/learner-group.js:    return new Ember.RSVP.Promise(function(resolve) {
addon/models/learner-group.js:    return new Ember.RSVP.Promise(function(resolve) {
addon/models/learner-group.js:    return new Ember.RSVP.Promise(
addon/models/instructor-group.js:const { Promise, map } = RSVP;
addon/models/objective.js:    return new RSVP.Promise(resolve => {
addon/models/program-year.js:const { Promise } = RSVP;
addon/models/cohort.js:const { Promise } = RSVP;
addon/models/cohort.js:    return new Ember.RSVP.Promise(function(resolve) {
addon/models/cohort.js:    return new Ember.RSVP.Promise(function(resolve) {
addon/models/user.js:const { computed, PromiseProxyMixin, Object:EmberObject, RSVP } = Ember;
addon/models/user.js:const { all, map, Promise } = RSVP;
addon/authenticators/ilios-jwt.js:    @return {Ember.RSVP.Promise} A promise that resolves when an auth token is
addon/authenticators/ilios-jwt.js:    return new Ember.RSVP.Promise((resolve, reject) => {
addon/authenticators/ilios-jwt.js:   * @return {Ember.RSVP.Promise} The result of the request.
addon/services/current-user.js:const { hash, map, Promise } = RSVP;
addon/components/dashboard-calendar.js:const { all, map, Promise } = RSVP;
addon/components/single-event.js:const { Promise, map } = RSVP;
[...]

PR #100, opened 2017-10-26
Use async/await syntax in Instructor-group model

stopfstedt:
refs #79

### DIFF
--- a/addon/models/instructor-group.js
+++ b/addon/models/instructor-group.js
@@ -15,8 +15,8 @@ export default Model.extend({
 
   coursesFromOfferings: computed('offerings.[]', async function () {
     const offerings = await this.get('offerings');
-    const courses = await map(offerings.toArray(), async ilmSession => {
-      const session = await ilmSession.get('session');
+    const courses = await map(offerings.toArray(), async offering => {
+      const session = await offering.get('session');
       return session.get('course');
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit fb11e380f7182198d8712c2cb83a8c77fa658cfb

https://deploy-preview-100--ilios-common.netlify.com

addon/models/instructor-group.js
```diff
@@ -1,60 +1,42 @@
...
+  title: attr('string'),
+  school: belongsTo('school', {async: true}),
+  learnerGroups: hasMany('learner-group', {async: true}),
+  ilmSessions: hasMany('ilm-session', {async: true}),
+  users: hasMany('user', {async: true}),
+  offerings: hasMany('offering', {async: true}),
+
+  coursesFromOfferings: computed('offerings.[]', async function () {
+    const offerings = await this.get('offerings');
+    const courses = await map(offerings.toArray(), async ilmSession => {
```
jrjohnson: Doesn't make a difference in the results but `ilmSession` here should be `offering`.

stopfstedt: yep, copy/paste for the _lose_. thanks for catching that. fixed.

PR #101, opened 2017-10-26
Use async/await syntax in Term model.

stopfstedt:
refs #79.

### DIFF
--- a/addon/models/term.js
+++ b/addon/models/term.js
@@ -1,9 +1,8 @@
 import DS from 'ember-data';
 import Ember from 'ember';
 
-const { computed, isEmpty, RSVP } =  Ember;
+const { computed, isEmpty } =  Ember;
 const { collect, sum, gte } = computed;
-const { Promise } = RSVP;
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 13dda283cda6bfd8be71fe44967d0530701440fe

https://deploy-preview-101--ilios-common.netlify.com

addon/models/term.js
```diff
@@ -78,21 +69,16 @@ export default Model.extend({
...
-    });
+  allParentTitles: computed('parent.{title,allParentTitles.[]}', async function() {
+    const parentTerm = await this.get('parent');
+    if(!parentTerm){
+      return [];
+    }
+
+    const parents = await parentTerm.get('allParents');
+    const titles = parents.mapBy('title');
+    titles.push(this.get('parent.title'));
```
jrjohnson: Not sure if it matters, but since it is already in scope shouldn't this be `titles.push(parentTerm.get('title'));`?

stopfstedt: good catch. handled.

PR #124, opened 2017-12-01
Remove jwt-decode shim

jrjohnson:
This shim probably isn't going to be maintained for ember 2.17+ as it doesn't
seem to work. We can replace it with a simple implementation of our own.

This unlocks #123


jrjohnson:
Deploy preview ready!

Built with commit b67c493c9693a953a4516c0139a7065ba5e84c9c

https://deploy-preview-124--ilios-common.netlify.com

PR #128, opened 2017-12-06
Cohorts do not keep track of primary users

jrjohnson:
This is mainly a fix so that mirage will correctly understand that there
is no inverse property for this relationship on the cohort model.


jrjohnson:
Deploy preview ready!

Built with commit 97195514946075ee75b84cd094a60eb832d535d0

https://deploy-preview-128--ilios-common.netlify.com

PR #130, opened 2017-12-06
Convert promise result to array

jrjohnson:
Tacking .toArray() onto a promise in an await call doesn't call
`toArray` on the result of the promise. It calls it on the promise
itself and doesn't work.

### DIFF
--- a/addon/models/learner-group.js
+++ b/addon/models/learner-group.js
@@ -81,10 +81,10 @@ export default Model.extend({
     let users = await this.get('users');
     let subgroups = await this.get('children');
     let usersInSubgroups = await all(subgroups.mapBy('allDescendantUsers'));
-    let allUsers = (usersInSubgroups.reduce((array, set) => {
-      array.pushObjects(set);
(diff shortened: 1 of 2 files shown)


jrjohnson:
Deploy preview ready!

Built with commit 74dcf74449e97df21fe9f2137807924703a4d44f

https://deploy-preview-130--ilios-common.netlify.com

addon/models/learner-group.js
```diff
@@ -144,7 +145,7 @@ export default Model.extend({
   allDescendants: computed('children.@each.allDescendants', async function(){
     const descendants = [];
     const children = await this.get('children');
-    descendants.pushObjects(children.toArray());
```
stopfstedt: i don't understand this change. `this.get('children')` should a return a PromiseArray, so calling `toArray()` on it seems necessary here.

jrjohnson: Thanks, I got a little bit overzealous with my first replacement. I thought I had put them all back, but I missed this one.

PR #138, opened 2017-12-10
Converted Session model to async/await syntax

stopfstedt:
refs #79.

this also contains a bugfix to the `tmaxSingleOfferingDuration` CP, which is used in the frontend in various places (I can file individual tickets there for reference if need be).


jrjohnson:
Deploy preview ready!

Built with commit ae48a2fdc4c1fdf5b68058fd7e6c719169b08323

https://deploy-preview-138--ilios-common.netlify.com

PR #156, opened 2018-01-12
Use the API server to access the ICS feed

jrjohnson:
This is where the feed is, getting it from the current host was never a
good idea.


jrjohnson:

Built with commit 73da7ce180f4137755a8e01d2d9a4523dfc73e59

https://deploy-preview-156--ilios-common.netlify.com

PR #162, opened 2018-01-24
If no apiHost is configured use the current URL

jrjohnson:
When we're serving the frontend from an API server we don't actually
send and apiHost because it is just /. We need to handle this as a case
here otherwise the ICS feed URL won't be an absolute path.

Fixes ilios/frontend#3485


jrjohnson:

Built with commit 56ac55e9bb39991ec0483e2712f15a572c44346f

https://deploy-preview-162--ilios-common.netlify.com

PR #189, opened 2018-03-26
Fix some issues with mirage

jrjohnson:
1) Mirage should not intercept requests when in development otherwise the dummy app can't connect to the API.
2) Remove scenarios we never use


jrjohnson:

Built with commit 503bff85dc2981023760c6f42f619c1aeac920a9

https://deploy-preview-189--ilios-common.netlify.com

PR #194, opened 2018-04-10
Add methods to get descendant data from term model

jrjohnson:
We need the list of titles so we can do good filtering on the frontend


jrjohnson:

Built with commit b095f3693c7888a250f56a7af2466e592a601395

https://deploy-preview-194--ilios-common.netlify.com

PR #202, opened 2018-04-19
Use the model for checking permissions

jrjohnson:
Using the ID matches our backend setup, but there is no need for it here
and it unnecessarily complicates testing or requires looking things up
in the store that we probably already have elsewhere.


jrjohnson:

Built with commit 1d5e02e32cccc13f05480838cbcdc2792ff4b5a5

https://deploy-preview-202--ilios-common.netlify.com

PR #209, opened 2018-04-23
Make Update and Delete API consistent

jrjohnson:
Always pass the object we are checking even if the check is done on the
school


jrjohnson:

Built with commit 53a5558cc8ed1f07d9491e123438c83df96c892e

https://deploy-preview-209--ilios-common.netlify.com

PR #231, opened 2018-05-04
Avoid loading unnecessary data

jrjohnson:
By pre-selecting which roles we are actually searching for we can avoid
loading data that we don't need. This can speed up checking
significantly in some situations.


jrjohnson:

Built with commit 84a5d92d50db39f1c0a9b7f2dc26f5bdb53d0737

https://deploy-preview-231--ilios-common.netlify.com

PR #237, opened 2018-05-11
Replace ember-tooltips with ember-attachment

jrjohnson:
Tooltips is in transition and having trouble staying up to date. It is
kicking out a lot of warnings right now. Since ember-attachment does
everything we need we can just use it.


jrjohnson:

Built with commit e66accf71b683cfe9bb82fcfed0c12c5e9c8c0f8

https://deploy-preview-237--ilios-common.netlify.com

PR #246, opened 2018-06-01
eliminate callbacks for objectives/competencies in single event component

stopfstedt:
companion ticket to ilios/ilios#2173

### DIFF
--- a/addon/components/single-event.js
+++ b/addon/components/single-event.js
@@ -1,12 +1,9 @@
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import { computed } from '@ember/object';
-import RSVP from 'rsvp';
 import { isBlank, isEmpty } from '@ember/utils';
 import layout from '../templates/components/single-event';
 import moment from 'moment';
-
-const { map } = RSVP;
 
 export default Component.extend({
   layout,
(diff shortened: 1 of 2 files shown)


jrjohnson:

Built with commit 9bde0f006e480a9dea85af7310257b91d3d3fd33

https://deploy-preview-246--ilios-common.netlify.com

addon/components/single-event.js
```diff
@@ -168,4 +177,34 @@ export default Component.extend({
...
+  /**
+   * Callback function for <code>Array.sort()<code>.
+   * Compares two given Objects by their position property (in ascending order), and then by id (descending).
+   *
+   * @method positionSortingCallback
+   * @param {Object} obj1
+   * @param {Object} obj2
+   * @return {Number}
+   */
+  positionSortingCallback(obj1, obj2) {
```
jrjohnson: Why not use the `sortable-by-position` mixin here?

addon/components/single-event.js
```diff
@@ -99,31 +102,37 @@ export default Component.extend({
...
+      }
       return {
+        id: objective.id,
         title,
-        domain: competency.get('title') + ' (' + domain.get('title') + ')',
+        domain: competency.title + ' (' + domain.title + ')',
         position
       };
-    });
+    }).sort(this.positionSortingCallback);
```
jrjohnson: Just curious - how were we sorting this before? I don't see any where that was happening.

addon/components/single-event.js
stopfstedt: `sortable-by-position` mixin assumes Ember objects as input. we're operating on plain on javascript objects here, so `get()` won't work here. so i re-implemented that sorting callback to work with POJsOs.

addon/components/single-event.js
stopfstedt: we're using the output of the `sortedObjectives` CP before, which uses the callback provided by  `sortable-by-position` mixin.

addon/components/single-event.js
jrjohnson: Ah ok. Thanks.

PR #250, opened 2018-06-18
Remove old computed properties

jrjohnson:
We're not using these anymore now that everything we need is in the
event.


jrjohnson:

Built with commit edf4f528872af238c44a827521af88699fc123d6

https://deploy-preview-250--ilios-common.netlify.com

PR #255, opened 2018-07-09
skipping test due to weird borkage on Travis.

stopfstedt:


jrjohnson:

Built with commit 802eca6e11e1ddd2c3e08426b1638f6dc4bd091d

https://deploy-preview-255--ilios-common.netlify.com

tests/unit/models/competency-test.js
```diff
@@ -22,6 +22,8 @@ module('Unit | Model | Competency', function(hooks) {
     });
   });
 
+  // commented out until further notice. Travis/CI is currently rejecting this test for reasons unknown. [ST 2018/07/09]
+  /*
   test('domain', async function(assert) {
```
jrjohnson: Instead can you import `skip` from `qunit` and just replace `test` with `skip`. That way it will be counted as skipped, but not worry that it fails.

stopfstedt: d'oh! i knew that :|

PR #268, opened 2018-08-17
When an event is missing display an error message

jrjohnson:
A couple of changes make this possible.
1) The event properties no longer needed to be async as all the info is
available already. This removes the loader from the single-event
component.
2) I ported over the not-found component from the frontend so it can be
used here.


jrjohnson:

Built with commit c500d83ee325b78e2bcbad0541139871ca5fdb9e

https://deploy-preview-268--ilios-common.netlify.com

PR #269, opened 2018-08-20
Better definition for computed property

jrjohnson:
We need to specifically track this second level array, it works
correctly in ember data 3.2+ and the old definition did not.


jrjohnson:

Built with commit f86656f47ebd41daf994268e04d286ed8549644e

https://deploy-preview-269--ilios-common.netlify.com

PR #281, opened 2018-09-05
Make z-index reasonable

jrjohnson:
The 100 value makes inputs cut through modals. They are important, but
not that important. Set it to 5 instead.

Fixes ilios/frontend#4021


jrjohnson:

Built with commit 2d82266b65d96e540fee8754ed29b77892601397

https://deploy-preview-281--ilios-common.netlify.com

PR #318, opened 2018-10-02
froala is a dev dependency

stopfstedt:
exporting out Froala as dependency causes problems with configuration in the App that includes common.
so - make this a dev dependency instead.


jrjohnson:
Ok, this should be fixed in the next ember-froala release so we can both be on the lookout for, and update accordingly (moving this back where it belongs into `dependencies`).

jrjohnson:

Built with commit d9bf1faa39173fbcee0deaebb1f4c93ea5613e32

https://deploy-preview-318--ilios-common.netlify.com

PR #351, opened 2018-10-18
Fix the rebuilding of the addon-test-support directory

jrjohnson:
When this was included in ilios/frontend this hook gets called with no
tree, since we wouldn't actually want to modify the tree in that case we
can just move on.


jrjohnson:

Built with commit 08a05fc0efd7ed97ed975e6a29b24281e75cfeae

https://deploy-preview-351--ilios-common.netlify.com

PR #353, opened 2018-10-18
Skip canary tests

jrjohnson:
Ember Canary tests are slowing down our builds right now, they run
forever and don't pass.  Let's skip them for now.


thecoolestguy:
Merging, based on Jon's promise that "it works. I promise."

PR #357, opened 2018-10-22
Conditional link back to courses

### LINKED ISSUE #356: has-route template helper
we need to wrap our links to `courses` in a conditional check. that route may or may not be present in the parent app (frontend, lti, etc) 

write a template helper that deals with that.

stopfstedt:
fixes #356

### DIFF
--- a/addon/helpers/has-route.js
+++ b/addon/helpers/has-route.js
@@ -1,0 +1,11 @@
+import Helper from '@ember/component/helper';
+import { inject as service } from '@ember/service';
+
+export default Helper.extend({
+  routing: service('-routing'),
+
+  compute([value]) {
+    const routing = this.get('routing');
+    return routing.hasRoute(value);
+  },
+});
--- a/addon/helpers/has-route.js
+++ b/addon/helpers/has-route.js
@@ -1,11 +1,9 @@
 import Helper from '@ember/component/helper';
-import { inject as service } from '@ember/service';
+import { getOwner } from '@ember/application';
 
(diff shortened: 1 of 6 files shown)


addon/helpers/has-route.js
```diff
@@ -0,0 +1,11 @@
+import Helper from '@ember/component/helper';
+import { inject as service } from '@ember/service';
+
+export default Helper.extend({
+  routing: service('-routing'),
```
jrjohnson: There is now a public router service, but I don't know if it has this method:
https://www.emberjs.com/api/ember/release/classes/RouterService

stopfstedt: unfortunately not the same service. so i have to rely on the private service, still.
alternatively, routes can be looked up via `this.owner.lookup('route:' + routeName)`, but afaik this only works from within components.

jrjohnson:

Built with commit fd832defa3cae77e5bf6efcdfbde197566460c0b

https://deploy-preview-357--ilios-common.netlify.com

addon/helpers/has-route.js
jrjohnson: If you can look it up on the owner then you can use `import { getOwner } from '@ember/application';` to get that value. I've only used it in tests, but it should serve the same purpose in a helper. Probably worth it to avoid the private routing service lookup.

stopfstedt: fair point. i'll give that a try.

PR #364, opened 2018-10-23
ember-i18n to to ember-intl conversion

stopfstedt:
refs ilios/frontend#4077.


jrjohnson:

Built with commit 6a82f900554b749e7926ec764756a416de493ffe

https://deploy-preview-364--ilios-common.netlify.com

tests/unit/translations-test.js
```diff
@@ -1,32 +0,0 @@
-import { module, test } from 'qunit';
```
jrjohnson: This bit scares me. I don't see a solution though other than an open issue on the repo so I guess I'm just going to need to be more careful!

stopfstedt: alternatively, we could look into yaml specific diff tools and run them on travis out-of-band.

PR #412, opened 2018-12-14
Stop using jQuery to strip HTML tags

jrjohnson:
Ember is on a quest to remove jQuery, the less we do with it the better.


jrjohnson:

Built with commit 06de35e38712c15215865824166072b687bf7e50

https://deploy-preview-412--ilios-common.netlify.com

PR #413, opened 2018-12-17
moved remove-html-tags helper unit test from frontend.

stopfstedt:
the helper is in common now, but the corresponding test was still hanging around in frontend. 
move it here, where it belongs.


jrjohnson:

Built with commit 8eb5d96a752b3e73bae3e66f0f04e75a400364ac

https://deploy-preview-413--ilios-common.netlify.com

PR #426, opened 2019-01-09
Session linking in the Admin UI

stopfstedt:
refs ilios/frontend#1323

No postrequisite is selected, Due-date can be set if this is an ILM
selection_258

Postrequisite is selected, ILM due date cannot be set. The "Due Prior To" label links to the postrequisite session.
selection_254

Inline editing 
selection_257

List of linked/clickable prerequisites on the other end
selection_253

### DIFF
--- a/addon/components/session-overview.js
+++ b/addon/components/session-overview.js
@@ -256,11 +256,11 @@ export default Component.extend(Publishable, Validations, ValidationErrorDisplay
       this.set('postRequisite', postRequisite);
     },
 
-    changePostrequisite() {
+    async changePostrequisite() {
       const session = this.get('session');
       const postRequisite = this.get('postRequisite');
       session.set('postrequisite', postRequisite);
-      session.save();
+      await session.save();
     },
 
     async revertPostrequisiteChanges() {
--- a/addon/components/session-overview.js
+++ b/addon/components/session-overview.js
@@ -50,6 +50,7 @@ export default Component.extend(Publishable, Validations, ValidationErrorDisplay
   currentUser: service(),
   routing: service('-routing'),
   permissionChecker: service(),
+  intl: service(),
   init() {
     this._super(...arguments);
     this.set('sortTypes', ['title']);
@@ -69,6 +70,10 @@ export default Component.extend(Publishable, Validations, ValidationErrorDisplay
       this.set('sessionType', sessionType);
     });
 
+    this.get('session.postrequisite').then(postRequisite => {
+      this.set('postRequisite', postRequisite);
+    });
+
     this.get('session.sessionDescription').then(sessionDescription => {
       if (sessionDescription){
         this.set('description', sessionDescription.get('description'));
@@ -86,6 +91,7 @@ export default Component.extend(Publishable, Validations, ValidationErrorDisplay
   sortTypes: null,
   sessionTypes: null,
   sessionType: null,
+  postRequisite: null,
   sortedSessionTypes: sort('filteredSessionTypes', 'sortTypes'),
   showCheckLink: true,
   isSaving: false,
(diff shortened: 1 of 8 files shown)


jrjohnson:

Built with commit d8367221df2b6b636c4534a7d207fa50bc0c4c65

https://deploy-preview-426--ilios-common.netlify.com

saschaben:
This is looking great. I think now that we can see it that greying out, rather than striking out, is likely the better option for user display with the due date/due prior to labels, but if it is a significant harder effort, it can wait.

addon/components/session-overview.js
```diff
@@ -207,6 +244,30 @@ export default Component.extend(Publishable, Validations, ValidationErrorDisplay
...
+      const sessions = await course.get('sessions');
+      const postRequisite = sessions.findBy('id', id);
+      this.set('postRequisite', postRequisite);
+    },
+
+    changePostrequisite() {
+      const session = this.get('session');
+      const postRequisite = this.get('postRequisite');
+      session.set('postrequisite', postRequisite);
+      session.save();
```
jrjohnson: Should probably be `await session.save()`.

addon/mixins/session-route.js
```diff
@@ -1,4 +1,14 @@
 import Mixin from '@ember/object/mixin';
 
 export default Mixin.create({
+  async model(params) {
```
jrjohnson: I think this is the default magic ember model implementation and this code may be unnecessary.

addon/components/session-overview.js
```diff
@@ -99,6 +105,37 @@ export default Component.extend(Publishable, Validations, ValidationErrorDisplay
...
+   * @type {Ember.computed}
+   */
+  postRequisiteTitle: computed('session.postrequisite', async function(){
+    const intl = this.get('intl');
+    const session = this.get('session');
+    const postRequisite = await session.get('postrequisite');
+    if (postRequisite) {
+      return postRequisite.get('title');
+    } else {
+      return intl.t('general.none');
```
jrjohnson: This looks to be code that won't get called since you have a similar check in the template that also resolves `None`. Might be worth removing so the `intl` service isn't needed.

addon/templates/components/session-overview.hbs
```diff
@@ -242,8 +242,53 @@
...
+        {{else}}
+          {{t "general.none"}}
+        {{/if}}
+      {{/if}}
+    </div>
+    <div class="prerequisites block">
+      <label>{{t "general.prerequisites"}}:</label>
+      {{#if session.hasPrerequisites}}
+        <span>
+          {{#each (await session.prerequisites) as |prerequisite|~}}
```
jrjohnson: I don't know what the `~` does here - extra empty char?

addon/templates/components/session-overview.hbs
```diff
@@ -242,8 +242,53 @@
...
+            {{#link-to "session.index" (await prerequisite.course) prerequisite}}{{fa-icon  "external-link-square-alt"}} {{prerequisite.title}}{{/link-to~}}
+            {{#unless (eq prerequisite session.prerequisites.lastObject)}}, {{/unless}}
```
jrjohnson: I lied in the meeting and don't have a solution for the trailing comma. My idea was to just the `join` helper which wouldn't add a last comma, but then the results wouldn't be linkable.

addon/components/session-overview.js
stopfstedt: we need this for the clickable label of the in-place editor. it's in use here:
https://github.com/ilios/common/pull/426/files#diff-98585cecdc6f319195f9e7d47acd4757R257

addon/templates/components/session-overview.hbs
stopfstedt: the tilde is for whitespace control, this got ported over from handlebars. check it out!
https://handlebarsjs.com/expressions.html#whitespace-control

addon/components/session-overview.js
stopfstedt: handled.

addon/mixins/session-route.js
stopfstedt: you are correct. yanked it.

PR #428, opened 2019-01-10
Drop support for old node versions

jrjohnson:
This is our own addon so we don't need to conform to ember sugestions
here and can only support and test against the latest node LTS version.


.travis.yml
```diff
@@ -1,9 +1,7 @@
 ---
 language: node_js
 node_js:
-  # we recommend testing addons with the same minimum supported node version as Ember CLI
-  # so that your addon works for all apps
-  - "6"
+  - "10"
```
Trott: If you always want to get the latest LTS without having to update `.travis.yml` every October, use this:

```suggestion
  - "lts/*"
```

jrjohnson: I'd still need to update `package.json` though, unless there is a notation I can use there as well?

jrjohnson:

Built with commit 25ede03f40f34d185c057a1ef8803b1243b8571c

https://deploy-preview-428--ilios-common.netlify.com

PR #457, opened 2019-01-25
Filter events correctly

jrjohnson:
Bad && here, was filtering out both events with no postreqs as well as
all ILMs instead of just Ilms with postreqs.

There is testing which covers this, but the PR that broke did not rebase with those tests :(


jrjohnson:

Built with commit fe9daade03a38ff74eab182fc4e58c939395c5d5

https://deploy-preview-457--ilios-common.netlify.com

PR #474, opened 2019-02-03
Improved Timezone Support

jrjohnson:
The timezone files shipped by moment are huge since they have to include so much information. Since we only support modern browsers we can drop the entire library and use a builtin which will save our users from downloading nearly 20% of the javascript we require!


jrjohnson:

Built with commit a11e952f5e8556539433ac7f5cb0a1bb65271d93

https://deploy-preview-474--ilios-common.netlify.com

PR #485, opened 2019-02-06
move slug generation functionality into mixin.

stopfstedt:
we need those slug-generating functions in the mixin since we want to splash it into other components, like the user-profile-calendar in frontend.

refs ilios/frontend#4290


jrjohnson:

Built with commit 8c64105469b83243f68d5fb47cb5e946515b0e17

https://deploy-preview-485--ilios-common.netlify.com

PR #491, opened 2019-02-08
fix cleanQuery() bug

Trott:
cleanQuery() only replaces the first special character. Fix it to
replace all special characters.

### DIFF
--- a/addon/utils/query-utils.js
+++ b/addon/utils/query-utils.js
@@ -3,5 +3,5 @@ import $ from 'jquery';
 const { trim } = $;
 
 export function cleanQuery(query) {
-  return trim(query).replace(/[-,?~!@#$%&*+'="]/, ' ');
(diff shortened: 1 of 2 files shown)


jrjohnson:

Built with commit 6c6555a1db6e65383a9dd5253fa611f3706ec0bd

https://deploy-preview-491--ilios-common.netlify.com

addon/utils/query-utils.js
```diff
@@ -2,6 +2,8 @@ import $ from 'jquery';
 
 const { trim } = $;
 
+// TODO: allow-list instead of block-list?
```
jrjohnson: This should go into an issue instead of a todo. It would require some discussion since I'm not sure we actually can whitelist.

Trott: OK, removed.

PR #495, opened 2019-02-11
Switch from ordinals to bullets

jrjohnson:
Ordinals were probably never a good idea here as it implies a ranking.
Let's use bullets!

Fixes ilios/frontend#4315


jrjohnson:

Built with commit 06635e1b9f1d20f1bf4577990eb056f70e9c7a30

https://deploy-preview-495--ilios-common.netlify.com

PR #514, opened 2019-02-13
Remove overridden promise

jrjohnson:
This is no longer necessary as the ember runloop has improved in 3.4 and should
handle our async/await much better now.


jrjohnson:

Built with commit 0f68246c96249d8b77bc5d70681f46c5444dd600

https://deploy-preview-514--ilios-common.netlify.com

PR #518, opened 2019-02-18
Reorganize the way the JWT is found

jrjohnson:
When JWT wasn't set (the user was logged out) this would die horribly.
Instead we need to do some validation steps and get the property in a
better way.

Fixes ilios/frontend#4291


jrjohnson:

Built with commit 267ed9088598991698fc3924b40fa72a55bf6765

https://deploy-preview-518--ilios-common.netlify.com

PR #528, opened 2019-02-20
Dedupe pre post requisites

stopfstedt:
refs ilios/frontend#4328

selection_313

selection_314

### DIFF
--- a/addon-test-support/ilios-common/page-objects/components/week-glance-pre-work.js
+++ b/addon-test-support/ilios-common/page-objects/components/week-glance-pre-work.js
@@ -1,6 +1,7 @@
 import {
   attribute,
   create,
+  isPresent,
   text
 } from 'ember-cli-page-object';
 
@@ -9,6 +10,8 @@ const definition = {
   title: text('[data-test-event-title]'),
   date: text('[data-test-date]'),
   url: attribute('href', '[data-test-date] a'),
+  moreInfo: text('[data-test-more]'),
+  hasMoreInfo: isPresent('[data-test-more]'),
 };
 
 export default definition;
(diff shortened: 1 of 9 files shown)


jrjohnson:

Built with commit 6c488666c72fed11351fc68e4201719886ace0b8

https://deploy-preview-528--ilios-common.netlify.com

jrjohnson:
I think at this early stage it can go without the <show all> button. Maybe `Due before ardvark (mm/dd/yy) and 14 later events.`?

stopfstedt:
> I think at this early stage it can go without the button. Maybe `Due before ardvark (mm/dd/yy) and 14 later events.`?

_later_ is not necessarily the case, they could be running at the same time. but i'll take it.

saschaben:
couple comments:
1) the "show all" button is sort of intrusive/ugly in this context. Maybe instead a clickable icon? bacon?
2) I think the wording we need is "with XX more offerings"
thoughts?

stopfstedt:
> couple comments:
> 
> 1. the "show all" button is sort of intrusive/ugly in this context. Maybe instead a clickable icon? bacon?
> 2. I think the wording we need is "with XX more offerings"
>    thoughts?

re: 1. yes on :bacon: !! srsly tho, i'm hearing "no expando" from @jrjohnson. what to do?
re: 2. agreed.

stopfstedt:
selection_315

next iteration, _no expando_.

@saschaben @jrjohnson any comments before i back this up with tests?

jrjohnson:
While I think we'll add expansion it isn't necessary now (IMO). I do think we need to express that the one you're seeing is the earliest version. Otherwise you will really want to see the rest to know how we picked.

PR #536, opened 2019-02-21
Add a null check when escaping regex

jrjohnson:
So we don't have to put a guard on every invocation of this.


jrjohnson:

Built with commit cfee5aec172fc3b76fdaadb1a5d09610810a1ef6

https://deploy-preview-536--ilios-common.netlify.com

PR #542, opened 2019-02-26
Move ember-data into dependencies

jrjohnson:
Let's provide this so apps don't have to manage it themselves.
Updated and hard coded it at the latest version since that is what we
were providing with the loose version before.


jrjohnson:

Built with commit 152e13ea766138bce21b4eb84e1d9c5a7cd72d31

https://deploy-preview-542--ilios-common.netlify.com

PR #567, opened 2019-03-15
rm published attrs from program and program-year.

stopfstedt:
refs ilios/frontend#3675 
refs ilios/frontend#4416

refs https://github.com/ilios/ilios/pull/2379


stopfstedt:
нет.
only flag those attributes as deprecated, for now.

PR #584, opened 2019-03-21
cosmetic changes to event details

djvoa12:
**Summary:**
Minor cosmetic wording updates as defined in the issue.

**UI:**
Screen Shot 2019-03-21 at 4 19 12 PM

Fixes https://github.com/ilios/frontend/issues/1378

### DIFF
--- a/addon/components/single-event.js
+++ b/addon/components/single-event.js
@@ -16,21 +16,19 @@ export default Component.extend({
   event: null,
   layout,
 
-  taughtBy: computed('event.instructors', 'intl.locale', {
-    get() {
-      const instructors = this.get('event.instructors');
-      if (isEmpty(instructors)) return '';
-      return this.get('intl').t('general.taughtBy', { instructors });
+  taughtBy: computed('event.instructors', 'intl.locale', function() {
+    const instructors = this.get('event.instructors');
+    if (isEmpty(instructors)) {
+      return '';
     }
-  }).readOnly(),
+    return this.get('intl').t('general.taughtBy', { instructors });
+  }),
 
-  sessionIs: computed('event.sessionType', 'intl.locale', {
-    get() {
-      const intl = this.get('intl');
-      const type = this.get('event.sessionTypeTitle');
-      return intl.t('general.sessionIs', { type });
-    }
-  }).readOnly(),
+  sessionIs: computed('event.sessionType', 'intl.locale', function() {
+    const intl = this.get('intl');
+    const type = this.get('event.sessionTypeTitle');
+    return intl.t('general.sessionIs', { type });
+  }),
 
   courseObjectives: computed('intl.locale', 'event.courseObjectives.[]', 'event.competencies.[]', function(){
     const intl = this.get('intl');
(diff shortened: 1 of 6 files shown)


addon/components/single-event.js
```diff
@@ -1,32 +1,36 @@
...
   'data-test-single-event': true,
+  event: null,
+  layout,
 
-  taughtBy: computed('intl.locale', 'event.instructors', function(){
-    const instructors = this.get('event.instructors');
-    if (isEmpty(instructors)) {
-      return '';
+  taughtBy: computed('event.instructors', 'intl.locale', {
+    get() {
```
jrjohnson: We're not using this pattern of having a getter anywhere even though it prevents writing to a compute property I find it hard to read and have removed it. Can you change that here?

addon/templates/components/single-event.hbs
```diff
@@ -89,33 +94,31 @@
     <div class="single-event-learningmaterial-list">
       <h2>{{t "general.sessionLearningMaterials"}}</h2>
       {{single-event-learningmaterial-list
-        learningMaterials=sessionLearningMaterials
-      }}
+        learningMaterials=sessionLearningMaterials}}
```
jrjohnson: Ember eslint used to prevent this, maybe they changed it, but we've got `}}` on it's own line everywhere else in a multi-line component. @stopfstedt how do you feel about this stylistically? Keep it on it's own line, or we don't care?

addon/components/single-event.js
```diff
@@ -1,32 +1,36 @@
...
+      const instructors = this.get('event.instructors');
+      if (isEmpty(instructors)) return '';
```
jrjohnson: I'll add linting to prevent this ASAP sorry, but single line if without curlys aren't our pattern so this should be changed to ```
if (isEmpty(instructors)) {
  return '';
}
```

File: addon/templates/components/single-event.hbs (hunk 2 above)
Comment (stopfstedt) [reply addon/templates/components/single-event.hbs]: i don't care one way or the other. _whatever eslint says it should be._

Comment (djvoa12) [reply addon/templates/components/single-event.hbs]: I used to have a new line for it then I started doing this style b/c it looked cleaner. If it's not an accepted thing here, I can revert and note this going forward.

File: addon/components/single-event.js (hunk 3 above)
Comment (djvoa12) [reply addon/components/single-event.js]: Darn. I really love single line `if` :(

File: addon/templates/components/single-event.hbs (hunk 2 above)
Comment (jrjohnson) [reply addon/templates/components/single-event.hbs]: Let's go with this for now, and I'll add linting and clean up everything later. It seems like we have a mix of both.

File: translations/fr.yaml [hunk 4]
```diff
@@ -229,7 +228,7 @@ general:
   selectVocabulary: "Choisi Vocabulaire"
   session: "Séance"
   sessionAdministration: "Administration de Séance"
-  sessionIs: "Cette séance est `{type}`"
+  sessionIs: "Type de Séance: `{type}`"
```
jrjohnson: Backticks should go away here in the same way quotes went away in the other translations.

jrjohnson:
@dartajax can you review and see how this looks?

PR #596, opened 2019-03-25
course learning materials sort and search additions

djvoa12:
**Summary:**
Course learning materials table is now FE sortable (with query params) and searchable.

**GIF:**

**Loading UI:**
Screen Shot 2019-03-25 at 1 48 57 PM

Accompanies https://github.com/ilios/frontend/pull/4450
Fixes https://github.com/ilios/frontend/issues/4402

### DIFF
--- a/addon/components/course-materials.js
+++ b/addon/components/course-materials.js
@@ -1,9 +1,12 @@
 import Component from '@ember/component';
 import EmberObject, { computed } from '@ember/object';
-import { or, reads } from '@ember/object/computed';
+import { reads } from '@ember/object/computed';
 import { isEmpty, isPresent } from '@ember/utils';
 import { all, task, timeout } from 'ember-concurrency';
 import layout from '../templates/components/course-materials';
+import DS from 'ember-data';
+
+const { PromiseArray } = DS;
 
 const DEBOUNCE_DELAY = 250;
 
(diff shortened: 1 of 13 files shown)


addon/components/loading-spinner.js
```diff
@@ -3,6 +3,10 @@ import layout from '../templates/components/loading-spinner';
 
 export default Component.extend({
   layout,
+
+  classNames: ['loading-spinner'],
   tagName: 'span',
-  classNames: ['loading-spinner']
+
+  class: '',
```
jrjohnson: This overloads this component too much and makes it harder to replace. Just stick with the regular spinner for now and don't worry about making it orange.

addon/mixins/course-materials-controller.js
```diff
@@ -1,6 +1,8 @@
 import Mixin from '@ember/object/mixin';
 
 export default Mixin.create({
-  queryParams: ['sortBy'],
-  sortBy: 'firstOfferingDate',
+  queryParams: ['clmSortBy', 'slmSortBy'],
```
jrjohnson: Names are a bit too obscure here. How about `courseSort` and `sessionSort`?

addon/mixins/sortable-table.js
```diff
@@ -1,19 +1,21 @@
...
   sortBy: '',
 
   sortedAscending: computed('sortBy', function(){
     const sortBy = this.get('sortBy');
     return sortBy.search(/desc/) === -1;
   }),
+
   actions: {
-    sortBy(what){
+    sortBy(prop) {
```
jrjohnson: Why make this change? It seems like unnecessary churn that adds to this diff. Should be a separate PR.

addon/templates/components/course-materials.hbs
```diff
@@ -1,117 +1,174 @@
...
+
+  {{#if (gt (get (await this.course.learningMaterials) "length") 0)}}
+    <span class="filter-course-lms">
+      <input
+        placeholder={{t "general.filterPlaceholder"}}
+        value={{this.clmQuery}}
+        oninput={{perform this.setClmQuery value="target.value"}}>
+    </span>
+
+    <table class="table__course">
```
jrjohnson: This class seems to only be used in tests. If it's for test selection you should use https://github.com/simplabs/ember-test-selectors which would look like `<table data-test-course-table>` .

addon/templates/components/course-materials.hbs
```diff
@@ -1,117 +1,174 @@
...
-        value={{filter}}
-        oninput={{action (mut filter) value="target.value"}}
         placeholder={{t "general.filterPlaceholder"}}
-      >
+        value={{this.slmQuery}}
+        oninput={{perform this.setSlmQuery value="target.value"}}>
     </span>
-    <table>
+
+    <table class="table__session">
```
jrjohnson: Should use test selectors here as well.

addon/components/course-materials.js
```diff
@@ -1,127 +1,152 @@
...
-import SortableTable from 'ilios-common/mixins/sortable-table';
+import { or, reads } from '@ember/object/computed';
+import { isEmpty, isPresent } from '@ember/utils';
+import { all, task, timeout } from 'ember-concurrency';
 import layout from '../templates/components/course-materials';
 
-export default Component.extend(SortableTable, {
+const DEBOUNCE_DELAY = 250;
+
+export default Component.extend({
```
jrjohnson: This component is really hard to understand. It was already in pretty bad shape and adding all of these additional things to it made it much more difficult. I think it would be significantly simplified by using only computed properties. Can you spend some time trying that and we can pair on it as well.

djvoa12:
@jrjohnson I addressed all your concerns with latest three commits.

addon/components/course-materials.js
```diff
@@ -1,126 +1,156 @@
...
+  sessionSort: null,
+  slmQuery: '',
+  onClmSort() {},
+  onSlmSort() {},
+  typesWithUrl: Object.freeze(['file', 'link']),
+
+  isClmLoading: reads('courseLearningMaterialObjects.isPending'),
+  isSlmLoading: reads('sessionLearningMaterialObjects.isPending'),
+
+  courseLearningMaterialObjects: computed(function() {
```
jrjohnson: Instead of returning a PromiseArray this should be written as ```
courseLearningMaterialObjects: computed(async function() {
    return this.fetchCourseLearningMaterials();
  }),
```

I'm also not sure why `fetchCourseLearningMaterials` has been extracted. Especially since that obscures that there should be dependent keys on `course` and `course.sessions.[]` and probably other things. Why not just include all of those in a single computed property?

File: addon/components/course-materials.js [hunk 8]
```diff
@@ -1,126 +1,156 @@
...
-              });
-            });
-          });
-        }).then(lmObjects => {
-          resolve(lmObjects);
-        });
-      });
-    });
+
+  sessionLearningMaterialObjects: computed(function() {
```
jrjohnson: Same comments as course above

addon/components/course-materials.js
```diff
@@ -1,126 +1,156 @@
...
-          let flat = learningMaterials.reduce((flattened, obj) => {
-            return flattened.pushObjects(obj.toArray());
-          }, []);
-
-          resolve(flat);
-        });
-      });
-    });
+
+  clmQuery: '',
```
jrjohnson: Since this is the external API of this component it would be clearer to define `clmQuery` and `slmQuery` as `courseQuery` in the same way as `courseSort`.

addon/components/course-materials.js
```diff
@@ -1,126 +1,156 @@
...
+  course: null,
+  courseSort: null,
+  sessionSort: null,
+  slmQuery: '',
+  onClmSort() {},
```
jrjohnson: Am I missing where these two methods are used? What is their purpose?

addon/components/course-materials.js
```diff
@@ -1,126 +1,156 @@
...
-                   (isPresent(obj.get('sessionTitle')) && obj.get('sessionTitle').match(exp));
-          });
-
-          resolve(filteredObjs);
-        }
-      });
-    });
+
+  filteredCourseLearningMaterialObjects: computed(
+    'courseLearningMaterialObjects', 'clmQuery', function() {
```
jrjohnson: `courseLearningMaterialObjects` returns a PromiseArray. I commented obove that it should return a promise. In either case this method should be async or depend on `courseLearningMaterialObjects.[]`.

addon/components/course-materials.js
```diff
@@ -1,126 +1,156 @@
...
+      firstOfferingDate,
+      sessionTitle: session.title,
+      title: lm.title,
+      type: lm.type,
+      url: lm.url
+    });
+  },
+
+  filterClmo(clmo) {
+    const exp = new RegExp(this.clmQuery, 'gi');
```
jrjohnson: This query will need to be escaped before it can be used here. There is a cleanQuery util for this.  I also seem to recall that @stopfstedt was using a non-regex pattern for doing this kind of filtering and can probably point you in that direction instead.

addon/components/course-materials.js
djvoa12: 👍

addon/components/course-materials.js
djvoa12: I thought about that. How would you get a loading state going while those async functions run?

addon/components/course-materials.js
djvoa12: The empty functions are used to signify functions are being passed in from parent component, just like a normal attribute. It's a common pattern.

Example:
```
hbs
{{child-component isLoading=isLoading onChange=(action "whatever")}}

js
isLoading: null,
onChange() {}
```

addon/components/course-materials.js
djvoa12: I just used what I extracted from below without changing anything.

https://github.com/ilios/common/blob/master/addon/components/course-materials.js#L67

I can look into it further if you feel that more is needed.

addon/components/course-materials.js
jrjohnson: You can use the [ember-promise-helpers](https://github.com/fivetanley/ember-promise-helpers) helpers for that.  In the template `{{#if (is-pending courseLearningMaterialObjects)}}` will give you a loading block and `{{#each (await courseLearningMaterialObjects) as |obj|}} will give you access to the promise results.

addon/components/course-materials.js
jrjohnson: What I'm not seeing is where these are actually called in this component. Did the name change?

addon/components/course-materials.js
jrjohnson: Yes, please change. Unescaped regex has caused us problems in the past if users enter a `/`. Pre-existing bug and not your fault, but worth cleaning up here.

addon/components/course-materials.js
djvoa12: I changed them to `onCourseSort` and `onSessionSort`.

addon/templates/components/course-materials.hbs
```diff
@@ -1,106 +1,153 @@
...
               {{/if}}
             </td>
-            <td>{{lmObject.type}}</td>
+            <td>{{capitalize lmObject.type}}</td>
             <td>{{lmObject.author}}</td>
           </tr>
+        {{else}}
+          <tr>
+            <td colspan="6" align="center">
+              {{if this.courseQuery "No Results Found" "No Course Learning Materials Available"}}
```
stopfstedt: Please pull this verbiage out and into translations.

addon/templates/components/course-materials.hbs
```diff
@@ -111,8 +158,14 @@
               {{/if}}
             </td>
           </tr>
+        {{else}}
+          <tr>
+            <td colspan="9" align="center">
+              {{if this.sessionQuery "No Results Found" "No Session Learning Materials Available"}}
```
stopfstedt: same here, these messages need to be translatable.

addon/templates/components/course-materials.hbs
djvoa12: Fixed.

addon/templates/components/course-materials.hbs
djvoa12: Fixed.

PR #597, opened 2019-03-25
update All Weeks link title

djvoa12:
**Summary:**
Updated `weeklyevents` link text from `Last Week / Next Week / All Weeks` to `All Weeks`.

**UI:**
Screen Shot 2019-03-25 at 6 03 00 PM

Fixes https://github.com/ilios/frontend/issues/3980

### DIFF
--- a/addon/templates/components/dashboard-week.hbs
+++ b/addon/templates/components/dashboard-week.hbs
@@ -1,16 +1,13 @@
 <div class="weeklylink" data-test-weekly-link>
   {{t "general.view"}}:
-  {{#link-to
-    "weeklyevents"
-    (query-params expanded=expanded)
-  }}
-    {{t "general.lastNextAllWeeks"}}
+  {{#link-to "weeklyevents" (query-params expanded=expanded)}}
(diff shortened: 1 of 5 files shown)


addon/templates/components/dashboard-week.hbs
```diff
@@ -1,16 +1,13 @@
 <div class="weeklylink" data-test-weekly-link>
   {{t "general.view"}}:
-  {{#link-to
-    "weeklyevents"
-    (query-params expanded=expanded)
-  }}
-    {{t "general.lastNextAllWeeks"}}
```
jrjohnson: Can you also remove `general.lastNextAllWeeks` from the translation files, I don't think we're using it anywhere.

djvoa12: Sure. Dave and I thought it might be useful to have it for future reference.

jrjohnson: makes sense, but we can add it back later. If we don't take it out now it will stick around forever.

PR #599, opened 2019-03-26
fixed buggy time inputs

djvoa12:
**Summary:**
Fixed buggy time inputs.
- Added `timeout` for better user experience; users would expect to be able to delete the 0
- `<input type="number>` is supported by all browsers (see [here](https://www.w3schools.com/tags/att_input_type_number.asp))
- All values entered are valid so no need for additional validations (non-number are automatically considered 0)

**GIF:**

Fixes https://github.com/ilios/frontend/issues/2049

### DIFF
--- a/addon/components/offering-form.js
+++ b/addon/components/offering-form.js
@@ -1,13 +1,14 @@
-import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import { computed } from '@ember/object';
+import { not } from '@ember/object/computed';
+import { inject as service } from '@ember/service';
+import { isEmpty, isPresent } from '@ember/utils';
 import {
   Promise as RSVPPromise,
-  map,
   filter,
-  hash
+  hash,
+  map
 } from 'rsvp';
-import { isEmpty, isPresent } from '@ember/utils';
 import moment from 'moment';
 import { validator, buildValidations } from 'ember-cp-validations';
 import ValidationErrorDisplay from 'ilios-common/mixins/validation-error-display';
--- a/addon/components/offering-form.js
+++ b/addon/components/offering-form.js
@@ -14,7 +14,7 @@ import ValidationErrorDisplay from 'ilios-common/mixins/validation-error-display
 import { task, timeout } from 'ember-concurrency';
 import layout from '../templates/components/offering-form';
 
-const { not } = computed;
+const DEBOUNCE_DELAY = 750;
 
 const Validations = buildValidations({
   room: [
@@ -39,14 +39,14 @@ const Validations = buildValidations({
       }),
     ]
   },
-  durationHours: [
+  hourBuffer: [
     validator('number', {
       allowString: true,
       integer: true,
       gte: 0
     })
   ],
-  durationMinutes: [
+  minuteBuffer: [
     validator('number', {
       allowString: true,
       integer: true,
@@ -63,14 +63,16 @@ const Validations = buildValidations({
         messageKey: 'general.smallGroupMessage'
       })
     ]
-  },
-
+  }
 });
 
 export default Component.extend(ValidationErrorDisplay, Validations, {
   currentUser: service(),
+
   layout,
+
   classNames: ['offering-form'],
+
   startDate: null,
   endDate: null,
   room: 'TBD',
@@ -93,6 +95,9 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
   recurringDayOptions: null,
   loaded: false,
   'data-test-offering-form': true,
+  hourBuffer: null,
+  minuteBuffer: null,
+
   associatedSchools: computed('cohorts.[]', function(){
     return new RSVPPromise(resolve => {
       const cohorts = this.get('cohorts');
@@ -402,6 +407,7 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
 
     this.setProperties({startDate, endDate, room, learnerGroups, recurringDays, instructors, instructorGroups, loaded});
   }).drop(),
+
   saveOffering: task(function * () {
     this.set('offeringsToSave', 0);
     this.set('savedOfferings', 0);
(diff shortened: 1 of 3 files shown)


addon/templates/components/offering-form.hbs
```diff
@@ -1,74 +1,78 @@
...
-              "has-error"
-            }}
-          >
-          {{#if
-            (and (v-get this "durationHours" "isInvalid") (is-in showErrorsFor "durationHours"))
-          }}
-            <span class="validation-error-message">{{v-get this "durationHours" "message"}}</span>
-          {{/if}}
-          <label>{{t "general.hours"}}</label>
+          <div class="hours__container">
```
jrjohnson: Prefer dash to double underscore so this would be `hours-container`.

addon/components/offering-form.js
```diff
@@ -439,28 +446,40 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
...
-  updateDurationMinutes: task(function * (minutes) {
-    let {validations} = yield this.validate();
-    this.send('addErrorDisplayFor', 'durationMinutes');
 
-    if (validations.get('durationMinutes.isInvalid')) {
+  updateDurationHours: task(function * (hours) {
+    yield timeout(DEBOUNCE_DELAY);
+    this.set('hourBuffer', hours);
+    const { validations } = yield this.validate();
+    const hourValidation = validations.toArray()[2];
```
jrjohnson: Where does the `[2]` come from? I'm worried it's just the second validation and would change if the ordering changed. If that's true it should be done by name and not index.

addon/components/offering-form.js
```diff
@@ -439,28 +446,40 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
 
     yield this.get('saveOffering').perform();
   }),
-  updateDurationMinutes: task(function * (minutes) {
-    let {validations} = yield this.validate();
-    this.send('addErrorDisplayFor', 'durationMinutes');
 
-    if (validations.get('durationMinutes.isInvalid')) {
+  updateDurationHours: task(function * (hours) {
+    yield timeout(DEBOUNCE_DELAY);
```
jrjohnson: What's the reason for a denouncing here? Since there isn't any async behavior I'm not sure why user input wouldn't be reflected immediately.

addon/components/offering-form.js
```diff
@@ -439,28 +446,40 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
...
-  updateDurationHours: task(function * (hours) {
-    let {validations} = yield this.validate();
-    this.send('addErrorDisplayFor', 'durationHours');
 
-    if (validations.get('durationHours.isInvalid')) {
+  updateDurationMinutes: task(function * (minutes) {
+    yield timeout(DEBOUNCE_DELAY);
+    this.set('minuteBuffer', minutes);
+    const { validations } = yield this.validate();
+    const minValidation = validations.toArray()[3];
```
jrjohnson: Magic `[3]` here. Same as above, should be referenced by name.

addon/components/offering-form.js
```diff
@@ -439,28 +446,40 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
...
     yield this.get('saveOffering').perform();
   }),
-  updateDurationMinutes: task(function * (minutes) {
-    let {validations} = yield this.validate();
-    this.send('addErrorDisplayFor', 'durationMinutes');
 
-    if (validations.get('durationMinutes.isInvalid')) {
+  updateDurationHours: task(function * (hours) {
+    yield timeout(DEBOUNCE_DELAY);
+    this.set('hourBuffer', hours);
```
jrjohnson: I think what you're trying to do here is to validate the data without changing the value, the problem is that this invalidates the work done on save. This isn't a good solution if it doesn't allow validations to work the same in all tasks.

addon/templates/components/offering-form.hbs
djvoa12: 👍

addon/components/offering-form.js
djvoa12: I don't think that varies. I think it goes by the order of the validations written.

```
  room: [
   ... 
  ],
  numberOfWeeks: {
    ...
  },
  hourBuffer: [
    ...
  ],
  minuteBuffer: [
    ...
  ],
  learnerGroups: {
     ...
  }
```
This order would dictate the array order. I can look again but it wasn't an object when I checked.

jrjohnson: That's my concern. It varies if you add a validation or move them around.

addon/components/offering-form.js
djvoa12: I just checked. It's a non-issue as long as the validation check is working properly. When the validations were off, it was showing a value of 0 immediately after each entry so it was really confusing for users. So it was to remove the terrible UX.

djvoa12:
@stopfstedt no validation is needed. Users are physically unable to enter non-numeric characters or negative values. Those all result in an empty string value, which result to 0.

stopfstedt:
> @stopfstedt no validation is needed. Users are physically unable to enter non-numeric characters or negative values. Those all result in an empty string value, which result to 0.

worked for me, i entered `-190` into the minutes field. which triggered an incorrect conversion to "3 hours, 50 minutes".

in order to trigger it, i had to move the cursor _before_ then `0` and then type really fast. no-one in their right mind will do that.

the other, way to trigger this is by backspace-deleting the `0` and then CTRL-V copy a negative value in. again, you'll have to be really fast. again, i don't see anyone doing that.

that being said, i'd rather err on the side of caution and have the validation in place. please add this back in, thanks!

Selection_408

dartajax:
I just ran this one by Tim G. He and I agree it is an improvement for sure. It still seems a bit weird (to me anyway) when you delete the zero and it comes back in a second or so. Then you type and a second after that the screen re-calculates everything.

PR #602, opened 2019-03-27
updated session table first offering column display control flow

djvoa12:
**Summary:**
Updated session list first offering column display control flow.

**UI:**
Screen Shot 2019-03-28 at 6 56 28 PM

Fixes https://github.com/ilios/frontend/issues/4298


djvoa12:
@dartajax I need help verifying all the control flow conditions work as intended.

dartajax:
This is what I'm getting ...

image

dartajax:
http://localhost:4200/courses/1573

dartajax:
image

dartajax:
LGTM 👍

addon/helpers/truncate-text.js
```diff
@@ -0,0 +1,9 @@
+import { helper } from '@ember/component/helper';
+
+export function truncateText([input], hash = {}) {
```
jrjohnson: Instead of creating a helper you can use `truncate` from [ember-string-helpers](https://github.com/romulomachado/ember-cli-string-helpers#truncate)
Like: https://github.com/ilios/common/blob/7553a10c3f8b7cd6a291a8ae316f9ff244185065/addon/templates/components/sessions-grid-offering.hbs#L60-L63

addon/templates/components/sessions-grid.hbs
```diff
@@ -30,65 +27,85 @@
       <span class="session-grid-first-offering">
         {{#if (await obj.session.firstOfferingDate)}}
           {{#if obj.isIlm}}
-            <strong>{{t "general.ilm"}}: {{t "general.dueBy"}}</strong> {{moment-format (await obj.session.firstOfferingDate) "L"}}
+            {{#if obj.postrequisite}}
+              <b>
```
jrjohnson: My reading of https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b#Usage_notes is that `strong` is preferred to `b`. What do you think?

addon/templates/components/sessions-grid.hbs
```diff
@@ -30,65 +27,85 @@
...
     {{#if (contains obj.session.id confirmDeleteSessionIds)}}
       <div class="confirm-removal" data-test-confirm-removal>
         {{t "general.confirmSessionRemoval"}}
         <div class="confirm-buttons">
-          <button {{action (perform removeSession obj.session)}} class="cancel" disabled={{removeSession.isRunning}} data-test-yes>
+          <button
+            class="cancel"
+            data-test-yes
+            disabled={{removeSession.isRunning}}
+            onclick={{perform removeSession obj.session}}>
```
jrjohnson: If replacing `{{action}}` with `onclick` you also need to handle `onpress` and maybe other keyboard events. I think the change is good, but please investigate how to handle the user pressing `<Enter>` or `<Space>` on the kayboard while the button has focus and see what ember does automatically for us that now needs to be replaced.

app/styles/ilios-common/mixins/sessions-grid.scss
```diff
@@ -60,4 +60,8 @@
       color: $dark-grey;
     }
   }
+
+  .fa-user-clock {
+    font-size: 14px;
```
jrjohnson: Font sizes should be in `rem` units.

addon/helpers/truncate-text.js
djvoa12: 👍

app/styles/ilios-common/mixins/sessions-grid.scss
djvoa12: 👍

addon/templates/components/sessions-grid.hbs
djvoa12: I was reading about it yesterday and the difference seems to be trivial. I am fine with `strong` as it signifies importance.

addon/templates/components/sessions-grid.hbs
djvoa12: The action handler gets called properly when I focus on the button and then press `enter` or `space`. So no further changes needed here.

PR #623, opened 2019-04-12
preserve session-list scroll

djvoa12:
**Summary:**
Utilized vanilla JS and `preserve-scroll` service to preserve window position while scrolling the session list. If users are returning to session list from session details, their scroll position is preserved.

**UI:**

Accompanies https://github.com/ilios/frontend/pull/4491
Fixes https://github.com/ilios/frontend/issues/3813

### DIFF
--- a/addon/components/sessions-grid.js
+++ b/addon/components/sessions-grid.js
@@ -1,13 +1,21 @@
 import Component from '@ember/component';
-import { isEmpty } from '@ember/utils';
 import { computed } from '@ember/object';
+import { next } from '@ember/runloop';
+import { isEmpty } from '@ember/utils';
+import { inject as service } from '@ember/service';
 import escapeRegExp from '../utils/escape-reg-exp';
 import { task } from 'ember-concurrency';
 import layout from '../templates/components/sessions-grid';
+import { addEventListener, runDisposables } from 'ember-lifeline';
 
 export default Component.extend({
+  preserveScroll: service(),
+  router: service(),
+
   layout,
+
   classNames: ['sessions-grid'],
+
   'data-test-sessions-grid': true,
   confirmDeleteSessionIds: null,
 
@@ -46,7 +54,6 @@ export default Component.extend({
     if (sortInfo.descending) {
       return sessions.sortBy(sortInfo.column).reverse();
     }
-
     return sessions.sortBy(sortInfo.column);
   }),
 
@@ -64,6 +71,24 @@ export default Component.extend({
     this.set('confirmDeleteSessionIds', []);
   },
 
+  didInsertElement() {
+    this._super(...arguments);
+    this.scrollDown();
+  },
+
+  didRender() {
+    this._super(...arguments);
+    addEventListener(this, window, 'scroll', () => {
+      const isCourseRoute = this.router.currentRouteName === 'course.index';
+      const preserveScroll = this.preserveScroll;
+
+      if (isCourseRoute) {
+        const yPos = window.scrollY;
+        preserveScroll.set('yPos', yPos === 0 ? null : yPos);
+      }
+    });
+  },
+
   actions: {
     confirmDelete(sessionId) {
       this.confirmDeleteSessionIds.pushObject(sessionId);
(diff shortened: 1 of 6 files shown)


djvoa12:
This was a difficult problem to solve. I looked into numerous addons, such as [this](https://www.emberobserver.com/addons/ember-scrollable), but found that using in-house code and jQuery were the best way to go for the moment.

When to use the hook to scroll down was super tricky because the session list renders **after** the page renders so I could not even rely on `didRender` component hook. Thus, I used a delayed function to run after `sortedSessions` CP returned as want the list to be rendered before we can scroll down. Might not be the ideal solution but it was the best I could come up with at the moment and it works.

stopfstedt:
it's a clever solution. just not sure if it's worth bring jquery back for. @jrjohnson  thoughts?

jrjohnson:
I think we're going to be able to ditch jQuery in a few releases once ember-data finishes converting to the fetch API. So I'd rather not add it again. It looks like the only API you're using though is scroll right? Can we use https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll instead?

Or just use the existing [scroll-to](https://github.com/ilios/common/blob/master/addon/utils/scroll-to.js) utility and we'll clean this up when we remove jQuery from that.

djvoa12:
@stopfstedt @jrjohnson I refactored out of jQuery to vanilla JS. It works.

addon/components/sessions-grid.js
```diff
@@ -64,6 +73,22 @@ export default Component.extend({
     this.set('confirmDeleteSessionIds', []);
   },
 
+  didRender() {
+    this._super(...arguments);
+    run(() => {
+      window.addEventListener('scroll', () => {
```
jrjohnson: This event listener needs to be cleaned up when this component is removed. The best way to do that is probably using `ember-lifeline` which has an example at https://ember-lifeline.github.io/ember-lifeline/docs/add-event-listener

addon/components/sessions-grid.js
```diff
@@ -46,7 +53,9 @@ export default Component.extend({
     if (sortInfo.descending) {
       return sessions.sortBy(sortInfo.column).reverse();
     }
-
+    if (isPresent(sessions) && this.preserveScroll.isListenerOff) {
+      this.scrollDown();
```
jrjohnson: Why here and not in `didInsertElement`? The sessions are passed in synchonsouly so other than rendering there shouldn't be a delay, could you add this instead `didInsertElement` and instead a `Ember.run.next` block?

jrjohnson:
@djvoa12 this is a good solution, while keeping this PR open could you try another one? Instead of of tracking scroll position what if clicking `Back to Session List` included a query parameter for `sessionId` that could be used to scroll that session into view?

addon/components/sessions-grid.js
djvoa12: I vaguely recall in the previous versions of Ember, I was getting warnings in the console regarding setting a value during `didInsertElement`. They were recommending `didRender`. Do you recall?

djvoa12: Oh wait, I called in the function because the lifecycle hooks would be called after page rendered but before session-list was rendered. So scroll down would occur before session-list were visible.

djvoa12:
@jrjohnson the solution should work for both `Back to Session List` AND back button, no? I might be able to set a query-param on course controller before leaving the route to session.

addon/components/sessions-grid.js
djvoa12: 👍

addon/components/sessions-grid.js
```diff
@@ -64,6 +71,26 @@ export default Component.extend({
...
   },
 
+  didInsertElement() {
+    this._super(...arguments);
+    this.scrollDown();
+  },
+
+  didRender() {
+    this._super(...arguments);
+    runTask(this, () => {
```
jrjohnson: `runTask` with `runDisposibles`  is the wrong API here. You want to use the event listener API at https://ember-lifeline.github.io/ember-lifeline/docs/add-event-listener

djvoa12: I fixed it. I couldn't get `package-lock.json` to update still. Maybe I need help for that as I don't know what's going on.

jrjohnson:
@dartajax assigning this to you to test, and if it passes, merge.

djvoa12:
@jrjohnson moved code from frontend. @dartajax maybe a quick sanity-check that it works and merge?

dartajax:
I am having a few issues getting the changes and my local environment going on this one right now. I will try again tomorrow.

dartajax:
I tested it as well. Good stuff. It takes just a tiny bit of time but that positioning on the list will be quite nice for users.

PR #645, opened 2019-04-25
fixed expand text bug

djvoa12:
**Summary:**
The ellipses icon which normally expands the display text doesn't work because it's embedded in a element which also has a click event trigger and the `edit` task function gets hit first before bubbling the action down to the ellipses icon. The parent click event trigger is removed and moved to `big-text` component.

Fixes https://github.com/ilios/frontend/issues/2519

### DIFF
--- a/addon-test-support/ilios-common/page-objects/course.js
+++ b/addon-test-support/ilios-common/page-objects/course.js
@@ -31,8 +31,8 @@ export default create({
 
   header: {
     scope: '[data-test-course-header]',
-    title: text('.editable'),
-    edit: clickable('.clickable'),
+    title: text('[data-test-edit]'),
+    edit: clickable('[data-test-edit]'),
     set: fillable('input'),
     save: clickable('.done')
   },
@@ -48,7 +48,7 @@ export default create({
     externalId: {
       scope: '.courseexternalid',
       value: text('span', { at: 0}),
-      edit: clickable('.clickable'),
+      edit: clickable('[data-test-edit]'),
       set: fillable('input'),
       save: clickable('.done'),
       hasError: isVisible('.validation-error-message')
@@ -56,7 +56,7 @@ export default create({
     startDate: {
       scope: '.coursestartdate',
       value: text('span', { at: 0}),
-      edit: clickable('.clickable'),
+      edit: clickable('[data-test-edit]'),
       set: datePicker('input'),
       save: clickable('.done'),
       hasError: isVisible('.validation-error-message')
@@ -64,7 +64,7 @@ export default create({
     endDate: {
       scope: '.courseenddate',
       value: text('span', { at: 0}),
-      edit: clickable('.clickable'),
+      edit: clickable('[data-test-edit]'),
       set: datePicker('input'),
       save: clickable('.done'),
       hasError: isVisible('.validation-error-message')
@@ -72,7 +72,7 @@ export default create({
     level: {
       scope: '.courselevel',
       value: text('span', { at: 0}),
-      edit: clickable('.clickable'),
+      edit: clickable('[data-test-edit]'),
       set: fillable('select'),
       save: clickable('.done'),
       hasError: isVisible('.validation-error-message')
@@ -81,7 +81,7 @@ export default create({
     clerkshipType: {
       scope: '.clerkshiptype',
       value: text('span', { at: 0}),
-      edit: clickable('.clickable'),
+      edit: clickable('[data-test-edit]'),
       set: fillable('select'),
       save: clickable('.done')
     },
--- a/addon/templates/components/big-text.hbs
+++ b/addon/templates/components/big-text.hbs
@@ -2,7 +2,10 @@
   {{displayText}}
 </span>
 {{#if showIcons}}
-  <span {{action "expand"}} class="clickable" role="button">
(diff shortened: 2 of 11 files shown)


stopfstedt:
> This works, and it's an interesting solution. I would rather keep the bug in for now though than put in this bandaid. It's going to be too easy to break again later since filtering events by class name is totally unique in our code base.

i concur.

addon/templates/components/editable-field.hbs
```diff
@@ -1,11 +1,14 @@
 <span class="content">
   {{#unless isEditing}}
-    <span onclick={{perform edit}} class="clickable editable">
+    <span class="clickable editable">
       {{#if value}}
         {{#if looksEmpty}}
           {{fa-icon "edit" class="enabled"}}
```
jrjohnson: This also needs to respond to click `{{fa-icon "edit" class="enabled" click=(perform edit}}`.

djvoa12: Added.

addon-test-support/ilios-common/page-objects/course.js
```diff
@@ -32,7 +32,7 @@ export default create({
   header: {
     scope: '[data-test-course-header]',
     title: text('.editable'),
-    edit: clickable('.clickable'),
```
jrjohnson: Class shouldn't be used for test selectors anymore. This is a holdout from an old way. Please change to `[data-test-edit]`.

djvoa12: Fixed.

addon/templates/components/big-text.hbs
```diff
@@ -1,8 +1,16 @@
-<span class={{if renderHtml "list-reset"}}>
+<span
+  class="editable {{if renderHtml "list-reset"}}"
```
jrjohnson: Is `editable` needed here now that `[data-test-edit]` is used for tests?

addon/templates/components/big-text.hbs
```diff
@@ -1,8 +1,16 @@
...
+  data-test-edit
+  role="button"
+  onclick={{action onEdit}}>
   {{displayText}}
 </span>
 {{#if showIcons}}
-  <span {{action "expand"}} class="clickable" role="button">
+  <span
+    class="clickable editable"
```
jrjohnson: Is `editable` needed here now that `[data-test-edit]` is used for tests?

addon/templates/components/editable-field.hbs
```diff
@@ -1,14 +1,26 @@
 <span class="content">
   {{#unless isEditing}}
-    <span onclick={{perform edit}} class="clickable editable">
+    <span class="clickable">
```
jrjohnson: This span is no longer `clickable` since the action has moved down.

addon/templates/components/editable-field.hbs
```diff
@@ -1,14 +1,26 @@
...
       {{#if value}}
         {{#if looksEmpty}}
-          {{fa-icon "edit" class="enabled"}}
+          {{fa-icon "edit"
+            class="editable enabled"
```
jrjohnson: Is `editable` needed here now that `[data-test-edit]` is used for tests?

Should now be `clickable`.

addon/templates/components/editable-field.hbs
```diff
@@ -1,14 +1,26 @@
...
-          {{big-text text=value renderHtml=renderHtml}}
+          {{big-text
+            renderHtml=renderHtml
+            text=value
+            onEdit=(perform edit)}}
         {{/if}}
       {{else}}
-        {{clickPrompt}}
+        <span
+          class="editable"
```
jrjohnson: Is `editable` needed here now that `[data-test-edit]` is used for tests?

addon/templates/components/big-text.hbs
djvoa12: Once I removed `editable`, I think bunch of other related tests were breaking. I could further fix those if you want.

jrjohnson: Editable wasn't part of this element originally, so if other tests are breaking when it is removed then they should be fixed as well.

djvoa12: I can look into it. `editable` class was part of the overall `span` before I refactored it so other tests were still depending on that class.

djvoa12: Fixed.

jrjohnson: I still see `editable` here.

djvoa12: Done.

addon/templates/components/big-text.hbs
```diff
@@ -1,8 +1,16 @@
-<span class={{if renderHtml "list-reset"}}>
+<span
+  class={{if renderHtml "list-reset"}}
```
jrjohnson: missing `class='clickable'` which is required in some browsers to take the click on a span. These should really be buttons, but that's for another day.

addon/templates/components/editable-field.hbs
jrjohnson: The other now clickable things will need this class applied or they may not work in some browsers.

Issue #649, opened 2019-04-26
Filters Not Working On Student Calendar


dartajax:
Logged in as a student, I selected a Course to be displayed on the Calendar. The following screen shot shows the result of that action.

https://ilios3-demo.ucsf.edu/dashboard?academicYear=2018&show=calendar&showFilters=true

![image](https://user-images.githubusercontent.com/7493331/56829744-6fc28180-6819-11e9-87ff-9d0c822f84af.png)

djvoa12:
Technical findings:

After the user clicks the checkbox shown on the screen shot above, the following CP gets triggered:
```
  eventsWithSelectedCourses: computed('ourEvents.[]', 'selectedCourses.[]', async function(){
    const events = await this.ourEvents;
    const selectedCourses = (await this.selectedCourses).mapBy('id');
    if(isEmpty(selectedCourses)) {
      return events;
    }
    const matchingEvents = await map(events, async event => {
      if(event.ilmSession || event.offering) {
        const courseId = await this.get('userEvents').getCourseIdForEvent(event);
        if(selectedCourses.includes(courseId)) {
          return event;
        }
      }
      return null;
    });
    return matchingEvents.filter(event => {
      return ! isEmpty(event);
    });
  }),
```
It calls `getCourseIdForEvent`.
```
  async getCourseIdForEvent(event){
    const course = await this.getCourseForEvent(event);
    return course.get('id');
  },
```
Which calls `getCourseForEvent`.
```
  async getCourseForEvent(event){
    const session = await this.getSessionForEvent(event);
    return await session.get('course');
  },
```
Which calls `getSessionForEvent`.
```
  async getSessionForEvent(event){
    let intermediary;
    if(event.offering){
      intermediary = await this.get('store').findRecord('offering', event.offering);
    } else {
      intermediary = await this.get('store').findRecord('ilmSession', event.ilmSession);
    }
    return await intermediary.get('session');
  },
```
Narrowing down the issue to `this.get('store').findRecord('offering', event.offering);`

PR #650, opened 2019-04-26
archive icon links to access learning materials

djvoa12:
**Summary:**
Added archive linked icon (tooltip included) next to "Session Learning Materials" header, which transitions them to the my materials page. Only students can see this icon.

**UI:**
Screen Shot 2019-05-02 at 5 56 19 PM

Accompanies https://github.com/ilios/frontend/pull/4540
Fixes https://github.com/ilios/frontend/issues/2301

### DIFF
--- a/addon/routes/mymaterials.js
+++ b/addon/routes/mymaterials.js
@@ -1,0 +1,28 @@
+import Route from '@ember/routing/route';
+import { reads } from '@ember/object/computed';
+import { inject as service } from '@ember/service';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+import { task } from 'ember-concurrency';
+
+export default Route.extend(AuthenticatedRouteMixin, {
+  commonAjax: service(),
+  currentUser: service(),
+  iliosConfig: service(),
+
+  host: reads('iliosConfig.apiHost'),
+  namespace: reads('iliosConfig.apiNameSpace'),
+
+  model() {
+    return this.fetchModelData.perform();
+  },
+
+  fetchModelData: task(function* () {
+    const commonAjax = this.commonAjax;
+    const host = this.host;
+    const namespace = this.namespace;
+    const user = yield this.currentUser.model;
+    const url = `${host}/${namespace}/usermaterials/${user.id}`;
+    const data = yield commonAjax.request(url);
+    return data.userMaterials;
+  }).cancelOn('deactivate')
+});
--- a/addon/routes/mymaterials.js
+++ b/addon/routes/mymaterials.js
@@ -2,7 +2,6 @@ import Route from '@ember/routing/route';
 import { reads } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
-import { task } from 'ember-concurrency';
 
 export default Route.extend(AuthenticatedRouteMixin, {
   commonAjax: service(),
@@ -12,17 +11,13 @@ export default Route.extend(AuthenticatedRouteMixin, {
   host: reads('iliosConfig.apiHost'),
   namespace: reads('iliosConfig.apiNameSpace'),
 
-  model() {
-    return this.fetchModelData.perform();
-  },
-
-  fetchModelData: task(function* () {
+  async model() {
     const commonAjax = this.commonAjax;
     const host = this.host;
     const namespace = this.namespace;
-    const user = yield this.currentUser.model;
+    const user = await this.currentUser.model;
     const url = `${host}/${namespace}/usermaterials/${user.id}`;
-    const data = yield commonAjax.request(url);
+    const data = await commonAjax.request(url);
     return data.userMaterials;
-  }).cancelOn('deactivate')
+  }
 });
--- a/addon/templates/components/my-materials.hbs
+++ b/addon/templates/components/my-materials.hbs
@@ -1,0 +1,132 @@
+<div class="material-list">
+  {{#if (gt (get materials "length") 0)}}
+    <span class="course-filter">
+      <select onchange={{action @setCourseIdFilter value="target.value"}}>
+        <option value="">{{t "general.allCourses"}}</option>
+        {{#each this.courses as |courseObj|}}
+          <option
+            selected={{is-equal courseObj.id @courseIdFilter}}
+            value={{courseObj.id}}>
+            {{courseObj.title}}
+          </option>
+        {{/each}}
+      </select>
+    </span>
+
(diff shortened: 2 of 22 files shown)


jrjohnson:
Code looks great. @dartajax please test with students who do not have instructor level permissions to see what this allows them to access and how difficult it is to get back into the dashboard from here given their lack of navigation. We also need to see if giving students a view like this into a course causes lots of permissions issues if the learning materials are not visible or something isn't published.

dartajax:
OK. It'll take me a second to link it up. But I will test indeed.

dartajax:
I logged in as a student and the links didn't take me anywhere. However, logged in as admin user, it worked fine. Of course, in that situation the "Return To Courses" link is valid; whereas with a student logged in, the only navigable route for them would be to go back to the Calendar.

jrjohnson:
We'll need to link somewhere else then. That page is designed for admins and adding a bunch of options to it to make it work for students is going to make it very confusing. How about linking to `/mymaterials`? We could add some filtering there to allow a course to be selected.

dartajax:
Here is the reality on this one ...

* There should be only one link by Session Materials
* The link should go to ...
https://ilios-stage.ucsf.edu/mymaterials?course=[`Course_ID`]&sortBy=sessionTitle
* The link should only be available to students.

djvoa12:
@jrjohnson this PR has been updated to be in line with the new direction.

addon/templates/components/single-event.hbs
```diff
@@ -93,6 +93,12 @@
   <fieldset>
     <div class="single-event-learningmaterial-list">
       <h2>{{t "general.sessionLearningMaterials"}}</h2>
+      {{#if (and this.currentUser.userIsStudent (await this.courseId))}}
```
jrjohnson: `userIsStudent` returns a promise. Needs to be wrapped in an `await`.

app/styles/ilios-common/components/single-event.scss
```diff
@@ -8,6 +8,7 @@
 
   h2 {
     @include ilios-heading-h4;
+    display: inline;
```
jrjohnson: Instead I think the icon should go inside the `h2`, otherwise it's just hanging out on it's own and will break weirdly on smaller screens.

djvoa12: 👍

addon/templates/components/single-event.hbs
djvoa12: 👍

addon/templates/components/my-materials.hbs
```diff
@@ -0,0 +1,132 @@
...
+            {{t "general.course"}}
+          {{/sortable-th}}
+          {{#sortable-th
+            colspan=3
+            sortedAscending=this.sortedAscending
+            sortedBy=(or (eq @sortBy "title") (eq @sortBy "title:desc"))
+            click=(action "sortBy" "title")}}
+            {{t "general.title"}}
+          {{/sortable-th}}
+          <th class="hide-from-small-screen">{{t "general.instructor"}}</th>
```
jrjohnson: The translations for `instructor` doesn't seem to be in common yet.

addon/routes/mymaterials.js
```diff
@@ -0,0 +1,28 @@
...
+  iliosConfig: service(),
+
+  host: reads('iliosConfig.apiHost'),
+  namespace: reads('iliosConfig.apiNameSpace'),
+
+  model() {
+    return this.fetchModelData.perform();
+  },
+
+  fetchModelData: task(function* () {
```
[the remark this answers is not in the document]
djvoa12: That's how many people pull route data via ember-concurrency?

I could restore the below code if you want.
```
   const commonAjax = this.commonAjax;
    const host = this.host;
    const namespace = this.namespace;
    return new Promise(resolve => {
      this.currentUser.get('model').then(user => {
        const userId = user.get('id');
        let url = `${host}/${namespace}/usermaterials/${userId}`;
        commonAjax.request(url).then(data => {
          resolve(data.userMaterials);
        });
      });
    });
```

app/styles/ilios-common/components/single-event.scss
```diff
@@ -9,6 +9,11 @@
   h2 {
     @include ilios-heading-h4;
     margin: 1em 0;
+
+    .fa-archive {
```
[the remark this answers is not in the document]
djvoa12: 👍

addon/templates/components/single-event.hbs
```diff
@@ -92,7 +92,15 @@
 
   <fieldset>
     <div class="single-event-learningmaterial-list">
-      <h2>{{t "general.sessionLearningMaterials"}}</h2>
+      <h2>
+        {{t "general.sessionLearningMaterials"}}
+        {{#if (and (await this.currentUser.userIsStudent) (await this.courseId))}}
+          {{fa-icon "archive"
+            class="clickable"
```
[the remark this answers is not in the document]
djvoa12: 👍

addon/routes/mymaterials.js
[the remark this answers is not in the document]
jrjohnson: I guess I just don't see the need here since we're not calling a `set` anywhere. For clarity you could rewrite with `async/await` on the model function.

addon/templates/components/my-materials.hbs
djvoa12: I found it.

https://github.com/ilios/common/blob/master/translations/en-us.yaml#L124

jrjohnson: That's plural instructor(s). You'll need https://github.com/ilios/frontend/blob/master/translations/en-us.yaml#L141

app/styles/ilios-common/components/single-event.scss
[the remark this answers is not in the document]
djvoa12: Still need `font-size: 1rem;` tho.

addon/templates/components/my-materials.hbs
djvoa12: Doh!

djvoa12:
@jrjohnson a reminder that research is needed. The event object does not contain course id. We need to fetch it. This could just be merged as is as the app itself won't break; it just won't show the icon.

The issue is related to https://github.com/ilios/common/issues/649.

jrjohnson:
This doesn't work at all for students without elevated privileges because they cannot access the session API endpoint and that is required to get to the course.  We'll need to add the course id to events before this will work.

Requires https://github.com/ilios/ilios/pull/2435

stopfstedt:
> I used a student token to log in to common and the link to `Access All Materials for this Course` didn't go anywhere.

this checks out now.

PR #661, opened 2019-05-06
pulls pre-work ILMs out of activities list into own display.

### LINKED ISSUE #633: Pull ILM's Linked to Session Out of Activities List (Similar to Calendar)
The ILM's that have been linked to Sessions need to be pulled out of the Activities list and placed in a Pre-Work zone such as ...

image

stopfstedt:
fixes #633

### DIFF
--- a/addon/components/dashboard-agenda.js
+++ b/addon/components/dashboard-agenda.js
@@ -8,6 +8,7 @@ export default Component.extend({
   userEvents: service(),
 
   layout,
+  areEventsSelectable: true,
   /**
    * Days in advance of the current date.
    * @property daysInAdvance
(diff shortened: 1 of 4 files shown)


stopfstedt:
coming up against this mismatch between pre-work in activities list VS calendar view

Selection_498

Selection_499

@saschaben does it make sense to "de-dupe" the pre-work in the activities list, or long-hand all of them out? pls advise, thanks.

stopfstedt:
changes as per our conversation on 5/6.

- separate panes for regular activities and ilm pre-work, both scrollable.
- "de-duped" pre-work.

screencapture-localhost-4200-dashboard-2019-05-08-11_33_54

addon/components/dashboard-agenda.js
```diff
@@ -8,6 +8,7 @@ export default Component.extend({
   userEvents: service(),
 
   layout,
+  areEventsSelectable: true,
```
jrjohnson: Not sure about this constant. What's the purpose?

stopfstedt: this controls whether pre-work events are linked in the UI, or not. the default is `false`, so i need to pass this down in order to turn it on.

stopfstedt:
> One question. Also the style feels a bit off with events under the pre-work, may need another title or else some padding, not sure. Maybe @dartajax will know

i assume you're referring to the new pre-work pane. agreed, this could use some better padding. anything else? pls keep in mind that style tweaks will also affect the calendar view in the dashboard, since we're using the same component there.

addon/components/dashboard-agenda.js
stopfstedt: @jrjohnson just verified that this flag is not tied to permissions, it's just a switch that has to be explicitly flipped on, on a per-component basis.

PR #672, opened 2019-05-13
Add displayName to user model

jrjohnson:
This is part of the API in v1.37

Requires https://github.com/ilios/ilios/pull/2440

### DIFF
--- a/addon/models/user.js
+++ b/addon/models/user.js
@@ -158,19 +158,21 @@ export default Model.extend({
     }
   ),
 
-  fullName: computed('firstName', 'middleName', 'lastName', function() {
-    const { firstName, middleName, lastName } = this.getProperties('firstName', 'middleName', 'lastName');
+  fullName: computed('firstName', 'middleName', 'lastName', 'displayName', function() {
+    if (this.displayName) {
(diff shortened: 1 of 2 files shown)


addon/models/user.js
```diff
@@ -157,19 +158,21 @@ export default Model.extend({
...
+    if (!this.firstName || !this.lastName) {
       return '';
     }
 
-    const middleInitial = middleName?middleName.charAt(0):false;
+    const middleInitial = this.middleName?this.middleName.charAt(0):false;
 
     if (middleInitial) {
-      return `${firstName} ${middleInitial}. ${lastName}`;
+      return `${this.firstName} ${middleInitial}. ${this.lastName}`;
```
stopfstedt: should be `${this.middleInitial}` then.

jrjohnson: This is *exactly* how I messed this up the first time 😛 `middleInitial` isn't a property like the other two, it's defined on 170 as `const middleInitial = this.middleName?this.middleName.charAt(0):false;`.

stopfstedt: i stand corrected! _approved_

PR #685, opened 2019-05-21
Mixins be gone

### LINKED ISSUE #680: un-mix controllers, templates and routes.
currently, routes and controllers that are used downstream in apps are implemented as mixins. turns out - that's not necessary, they can be exported as such. _who knew!_

see https://github.com/ilios/common/pull/650/files for reference.

stopfstedt:
fixes #680


stopfstedt:
too hacky, and never got the demo app/test coverage to fully run again. i'm binning this.

PR #705, opened 2019-06-03
last update indicator on sessions and offerings.

### LINKED ISSUE #704: expose "last_updated" values for session, offering
provide a timestamp identifying the last fingerprint update for any session and for any offering within a session. this is _not_ a comprehensive review target, just a KISS timestamp to flag staleness/activity. 

timestamp would display on each session detail header; on each session list expansion block; each offering list row; each offering manager page header.

stopfstedt:
fixes #704

### DIFF
--- a/addon-test-support/ilios-common/page-objects/session.js
+++ b/addon-test-support/ilios-common/page-objects/session.js
@@ -118,7 +118,8 @@ export default create({
       set: fillable('select'),
       save: clickable('.done'),
       hasError: isVisible('.validation-error-message')
-    }
+    },
+    lastUpdated: text('.last-update'),
   },
 
   leadershipCollapsed,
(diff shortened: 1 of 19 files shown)


stopfstedt:
work in progress - just looking for feedback on the visuals at this point. @saschaben, does this look acceptable?

Selection_554

Selection_553

stopfstedt:
@dartajax please click test this in the course sessions grid, and in the sessions overview. please note that the displayed timestamp is _not_ supposed to change as you modify a session (or one of it's associated data points, like title, mesh, objective), but should display an updated timestamp after _refreshing_ the page post-update. thanks!

addon-test-support/ilios-common/page-objects/session.js
```diff
@@ -118,6 +118,10 @@ export default create({
       set: fillable('select'),
       save: clickable('.done'),
       hasError: isVisible('.validation-error-message')
+    },
+    lastUpdated: {
```
jrjohnson: I'm not sure why this one doesn't match `lastUpdated: text('.slast-update')` above.

stopfstedt: good catch. fixed.

PR #897, opened 2019-10-07
Use quill in place of Froala editor

jrjohnson:
Open source and free, plus it doesn't use jquery.

Needs lots of testing. I'm occasionally seeing performance issues, especially when clicking in and out of the editor or pasting big blocks. This may be due to my development environment, but would like feedback from others.


jrjohnson:
Passing on this. We're going to re-up froala license as it works and is much faster than Quill.

PR #900, opened 2019-10-07
when opening new offering form, scroll to its bottom.

### LINKED ISSUE #899: Clicking to Add New Offering Should Navigate User Further Down the Screen
Part of ilios/frontend#4809 in a way.

Steps ...

* Go to somewhere like this ... 
https://ilios3-demo.ucsf.edu/courses/1657/sessions/49971

* image

User should be routed to the bottom of the screen to add the offering details ...

image

Not just to ...

image

stopfstedt:
fixes #899

### DIFF
--- a/addon/components/new-offering.js
+++ b/addon/components/new-offering.js
@@ -1,7 +1,6 @@
 import { inject as service } from '@ember/service';
 import Component from '@ember/component';
 import layout from '../templates/components/new-offering';
-import scrollIntoView from 'scroll-into-view';
 
 export default Component.extend({
   store: service(),
@@ -11,10 +10,6 @@ export default Component.extend({
   courseStartDate: null,
   courseEndDate: null,
   smallGroupMode: true,
-  didRender() {
-    this._super(...arguments);
-    scrollIntoView(this.element.querySelector('.offering-form .buttons'));
-  },
   actions: {
     save(startDate, endDate, room, learnerGroups, instructorGroups, instructors){
       const store = this.get('store');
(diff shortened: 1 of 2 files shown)


addon/components/new-offering.js
```diff
@@ -10,6 +11,10 @@ export default Component.extend({
   courseStartDate: null,
   courseEndDate: null,
   smallGroupMode: true,
+  didRender() {
+    this._super(...arguments);
+    scrollTo('.offering-form .buttons', 1000);
```
jrjohnson: `scrollTo` is kind of a hack and will have problems with Octane. Instead use this directly with 
```
import scrollIntoView from 'scroll-into-view';
scrollIntoView(this.element);
```

stopfstedt:
ugh, problems. hanging this off of `didRender()` will cause scrolling down on subsequent re-renderings of this component, e.g. when you toggle from single offering to small groups.
super annoying. 
the better event would be `didInsertElement`, but there's no guarantee that the sub-component we're scrolling down to is in the dom just yet.
i'm taking this back to the bench...

jrjohnson:
Ok. This is a *super* good use of the did-insert octane modifier. So we probably just need to wait a week or two and this might be very easy. I actually think we should consider creating a `scroll-to` modifier if noone else has yet and publish it.

stopfstedt:
ok, cool. if the modifier doesn't pan out, or doesn't land in a timely fashion, then my latest iteration on this seems to check out too.

jrjohnson:
This is unlocked now with our current ember version. Documentation at https://emberjs.github.io/rfcs/0415-render-element-modifiers.html#reusable-helpers

Will need to add the https://github.com/emberjs/ember-render-modifiers as well.

stopfstedt:
superseded by #1245

PR #904, opened 2019-10-07
display instructor group members.

### LINKED ISSUE #903: Need Way To Expand / Show All Instructors in Instructor Group at Offering Level
https://ilios3-demo.ucsf.edu/courses/1547?filterBy=d1

image

It is possible (but not entirely obvious) to retrieve the Instructors involved with this offering by mousing over as shown below.

image

Now click to edit ... (the pencil icon) ...

From this lower detail level, there is no way to view / review the instructors who are members of this instructor group. The only way to see all the instructors is by the mouse-over event shown higher up in the ticket. We need to either display the members of the instructor group at the lower level, allow the ellipses to be clickable at the higher level. Plus if we reveal the names, we still should have a way of displaying which instructor group has been attached.

image

stopfstedt:
fixes #903 

**before**

Selection_059
Selection_060
Selection_061

**after**

Selection_056
Selection_055
Selection_058

### DIFF
--- a/addon-test-support/ilios-common/page-objects/components/instructor-selection-manager.js
+++ b/addon-test-support/ilios-common/page-objects/components/instructor-selection-manager.js
@@ -2,7 +2,7 @@ import {
   clickable,
   collection,
   fillable,
-  hasClass
+  hasClass, text,
 } from 'ember-cli-page-object';
 
 export default {
@@ -16,7 +16,11 @@ export default {
   instructors: collection('[data-test-instructors] li', {
     remove: clickable()
   }),
-  instructorGroups: collection('[data-test-instructor-groups] li', {
-    remove: clickable()
+  instructorGroups: collection('[data-test-instructor-group]', {
+    title: text('[data-test-instructor-group-title]'),
+    members: collection('[data-test-instructor-group-members] li', {
+      text: text(),
+    }),
+    remove: clickable('[data-test-instructor-group-title]')
   }),
 };
--- a/addon-test-support/ilios-common/page-objects/components/instructor-selection-manager.js
+++ b/addon-test-support/ilios-common/page-objects/components/instructor-selection-manager.js
@@ -19,9 +19,7 @@ export default {
   }),
   instructorGroups: collection('[data-test-instructor-group]', {
     title: text('[data-test-instructor-group-title]'),
-    members: collection('[data-test-instructor-group-members] li', {
-      text: text(),
-    }),
+    members: collection('[data-test-instructor-group-members] li'),
     remove: clickable('[data-test-instructor-group-title]')
   }),
 };
--- a/addon-test-support/ilios-common/page-objects/components/instructor-selection-manager.js
+++ b/addon-test-support/ilios-common/page-objects/components/instructor-selection-manager.js
@@ -2,7 +2,8 @@ import {
   clickable,
   collection,
   fillable,
-  hasClass, text,
+  hasClass,
+  text,
 } from 'ember-cli-page-object';
 
 export default {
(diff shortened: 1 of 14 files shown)


addon-test-support/ilios-common/page-objects/components/instructor-selection-manager.js
```diff
@@ -2,7 +2,7 @@ import {
   clickable,
   collection,
   fillable,
-  hasClass
+  hasClass, text,
```
jrjohnson: Newline needed.

addon-test-support/ilios-common/page-objects/components/instructor-selection-manager.js
```diff
@@ -16,7 +16,11 @@ export default {
   instructors: collection('[data-test-instructors] li', {
     remove: clickable()
   }),
-  instructorGroups: collection('[data-test-instructor-groups] li', {
-    remove: clickable()
+  instructorGroups: collection('[data-test-instructor-group]', {
+    title: text('[data-test-instructor-group-title]'),
+    members: collection('[data-test-instructor-group-members] li', {
+      text: text(),
```
jrjohnson: `text` is actually a default attribute or every page object once it is selected. If you just want the text of the entire element this can be omitted.

addon/components/detail-instructors-list.js
```diff
@@ -13,6 +14,14 @@ export default Component.extend({
   sortGroupsBy: null,
   sortedInstructors: sort('instructors', 'sortInstructorsBy'),
   sortedInstructorGroups: sort('instructorGroups', 'sortGroupsBy'),
+  instructorGroupMembers: computed('instructorGroups.[]', async function() {
+    const instructorGroups = await this.instructorGroups;
+    const groupsOfInstructors = await all(instructorGroups.mapBy('users'));
+    return groupsOfInstructors.reduce((array, set) => {
```
jrjohnson: I believe this can now be done with [Array.flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat). Not 100% certain though.

addon/templates/components/instructor-selection-manager.hbs
```diff
@@ -25,6 +12,33 @@
...
+
+{{#if (get (await instructorGroups) "length")}}
+  <h4>
+    {{t "general.instructorGroups"}}
+    ({{get (await instructorGroups) "length"}})
+  </h4>
+  <div data-test-instructor-groups>
+    {{#each (sort-by "title" (await instructorGroups)) as |instructorGroup|}}
+      <div class="instructor-group" data-test-instructor-group>
+        <span role="button" {{action "removeInstructorGroup" instructorGroup}} data-test-instructor-group-title>
```
jrjohnson: Can this just be a `<button>`?

addon-test-support/ilios-common/page-objects/components/instructor-selection-manager.js
stopfstedt: done.

addon-test-support/ilios-common/page-objects/components/instructor-selection-manager.js
stopfstedt: good to know. thanks! fixed.

addon/components/detail-instructors-list.js
stopfstedt: `flat()` works, but as it turns out those CPs are not even necessary, they were kruft left over from a previous interation of this component rework. removed.

addon/templates/components/instructor-selection-manager.hbs
stopfstedt: functionally, yes. but not style-wise. buttons would rewire more style overrides on top. i'd rather not.

Selection_062

saschaben:
LGTM. we might consider modifying the instructor group badge color from grey to something else, to keep in line with the existing patterns in learner groups of different color for label vs content, but this will work for me as-is.

PR #912, opened 2019-10-23
Add explicit this to components

jrjohnson:
Codemod plus manual cleanup off all instances where `this` can be used to denote a property on the component backing object.


dartajax:
I tested quite a bit. LGTM. 👍

PR #918, opened 2019-10-24
Get course from the event

### LINKED ISSUE #649: Filters Not Working On Student Calendar

jrjohnson:
The session is not available for students to access, but we have access
to the course ID here, we can just use that. This allows students to use
the course filtering options on the calendar.

Fixes #649

### DIFF
--- a/addon/mixins/events.js
+++ b/addon/mixins/events.js
@@ -25,9 +25,8 @@ export default Mixin.create({
    * @param {Object} event
    * @return {Promise.<Object>}
    */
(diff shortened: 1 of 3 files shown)


dartajax:
I tested using student token on demo3. Calendar filters now work as they should. 👍

PR #940, opened 2019-10-31
Improve sorting icons

jrjohnson:
Icons were backwards desc/asc I also replaced the descending sort icon
to one that is a bit clearer.

Fixes ilios/frontend#4909


dartajax:
I have tested this. LGTM. 💯 🥇 👍

PR #953, opened 2019-11-13
Update to Froala v3

jrjohnson:
Looks very nearly the same. Will need some user testing to detect any flaws in the new version.

Todo:
- [x] disable tracking beacon
- [x] figure out why license key is not working
- [x] correct z-indexing
- [x] remove current default link search options, and replace with https://www.ncbi.nlm.nih.gov/pubmed/


saschaben:
notes on update:

for link selection feature:
 - correct z-indexing
 - remove current default link search options, and replace with `https://www.ncbi.nlm.nih.gov/pubmed/`

dartajax:
image

jrjohnson:
Seems to be an issue on common and not a change in this PR. Merge away!

dartajax:
Font issue was a common issue not Froala upgrade issue.

PR #959, opened 2019-11-15
Improved UI for picking postrequisites

jrjohnson:
Work in progress:
- [x] Validate Approach
- [x] needs tests

Fixes ilios/frontend#4319


saschaben:
this is looking great. only comments are for minor layout issues (see below). I don't think we need sorting for this until we hear further from users whether that is necessary/desirable. good stuff!

Ilios Common Demo App 2019-11-20 15-06-39

PR #970, opened 2019-11-19
Replace import from Ember Data

jrjohnson:
These are the new import paths, code mod did the work and caught some other
stuff that seemed reasonable to leave in.


dartajax:
Did a quick functional review - looks good.

PR #976, opened 2019-11-21
Use fetch to make ember-data requests

jrjohnson:
Once we turn off jquery this will be the default, but for now I've
forced it on.

Changes for fetch based networking

With jQuery removed ember-data reverts to using fetch to get data, but
the simple auth mixin doesn't work for this so we have to set the
headers ourselves.


addon/adapters/ilios.js
```diff
@@ -1,15 +1,35 @@
...
+export default RESTAdapter.extend({
   iliosConfig: service(),
+  session: service(),
 
   host: reads('iliosConfig.apiHost'),
   namespace: reads('iliosConfig.apiNameSpace'),
 
+  /**
+   * Force Fetch since we're still using jQuery
+   * https://github.com/emberjs/data/issues/6759
```
stopfstedt: :+1:

PR #978, opened 2019-11-21
Add and use new TruncateText component

jrjohnson:
Meant to be a more functional replacement for `<BigText>` with better a11y
and improved ability to use it inside of `<EditableField>`.

Improved fix for ilios/frontend#2519

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,7 +10,10 @@ module.exports = {
 
   options: {
     babel: {
-      plugins: [ require.resolve('ember-auto-import/babel-plugin') ]
+      plugins: [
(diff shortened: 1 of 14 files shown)


index.js
```diff
@@ -10,7 +10,10 @@ module.exports = {
 
   options: {
     babel: {
-      plugins: [ require.resolve('ember-auto-import/babel-plugin') ]
+      plugins: [
+        require.resolve('ember-auto-import/babel-plugin'),
+        require.resolve('@babel/plugin-proposal-nullish-coalescing-operator'),
```
stopfstedt: "nullish" :|

PR #987, opened 2019-11-22
Add LearningMaterialUploader Component

jrjohnson:
The LM uploader has been working off of a jQuery based component and
addon in the frontend that was never migrated to common. It didn't fail
because we had the `ember-uploader` addon installed here already which
has the same component name, but was not implemented fully.

WIP:
- [x] confirm translations


app/styles/ilios-common/components/new-learningmaterial.scss
```diff
@@ -15,4 +15,29 @@
   .description {
     @include ilios-browser-defaults;
   }
+
```
stopfstedt: these new styles are inconsistent with the rest of the form. the label should sit above the control, not be to the right of it.

jrjohnson: I styled it the way it looks now in frontend. With the label above and the button by itself on a line it looked weird.

PR #991, opened 2019-11-22
i'd rather be glimmering

stopfstedt:
converts components to glimmer/octane syntax.

### DIFF
--- a/addon/components/week-glance-event.js
+++ b/addon/components/week-glance-event.js
@@ -1,10 +1,10 @@
-import Component from '@ember/component';
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
 
-export default Component.extend({
-  tagName: '',
-  actions: {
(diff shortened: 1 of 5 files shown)


addon/components/week-glance-event.js
```diff
@@ -1,10 +1,10 @@
...
 
-export default Component.extend({
-  tagName: '',
-  actions: {
-    sortString(a, b){
-      return a.localeCompare(b);
-    }
+export default class WeekGlanceEvent extends Component {
+
+  @action
```
jrjohnson: Unless I'm wrong you don't need `@action` here because you aren't calling `this` in the function. The purpose of `@action` seems to be to bind this. Some background in https://github.com/emberjs/rfcs/issues/547

stopfstedt: handled.

PR #995, opened 2019-11-23
Replace mixin with inheritance

jrjohnson:
Like OO and stuff. Allows user and school events services to be
converted to ES classes.


addon/components/single-event.js
```diff
@@ -3,10 +3,9 @@ import { computed } from '@ember/object';
 import { reads } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import { isBlank, isEmpty } from '@ember/utils';
-import EventMixin from 'ilios-common/mixins/events';
 import moment from 'moment';
 
-export default Component.extend(EventMixin, {
+export default Component.extend({
```
stopfstedt: just double-checking here -we're _not_ extending `EventsBase` on purpose here, right?

jrjohnson: That's correct. The method we were using from the mixin isn't here anymore and was replaced with `courseId: reads('event.course'),` so it can go away entirely.

stopfstedt: :+1:

PR #1023, opened 2019-12-05
Glimmerize course-visualization components

stopfstedt:


addon/components/visualizer-course-objectives.hbs
```diff
@@ -1,48 +1,55 @@
...
-      <chart.tooltip @title={{this.tooltipTitle}}>
-        {{this.tooltipContent}}
-      </chart.tooltip>
+<div
+  class="visualizer-course-objectives {{if (not @isIcon) "not-icon"}}"
+  {{did-insert (perform this.load) @course}}
+  {{did-update (perform this.load) @course}}
+>
+  {{#unless this.load.isRunning}}
+    {{#if (or true @isIcon this.objectiveWithMinutes.length)}}
```
jrjohnson: I'm confused by this repeating line. Seems like `(or true` will always be true. Maybe a testing artifact?

stopfstedt: yea i was wondering about those too. will see how removing this will affect the whole thing.

stopfstedt: yanked `true` from those conditional checks without repercussions.

PR #1039, opened 2019-12-12
Glimmer <CourseHeader>

jrjohnson:
Unwraps some previously shared publish menu functionality in order to avoid a mixin and creates a publish course component instead. Also fixes some layout problems with locked courses.

It turns out our existing brown scheduled course color doesn't have enough contrast, but I've allowed that to pass here anyway.

### DIFF
--- a/addon/components/course-header.js
+++ b/addon/components/course-header.js
@@ -1,57 +1,26 @@
-import Component from '@ember/component';
-import { computed } from '@ember/object';
-import { reject } from 'rsvp';
-import Publishable from 'ilios-common/mixins/publishable';
-import { validator, buildValidations } from 'ember-cp-validations';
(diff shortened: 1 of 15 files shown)


addon/components/course-header.js
```diff
@@ -10,6 +10,7 @@ export default class CourseHeaderComponent extends Component {
 
   @restartableTask
   *changeTitle() {
+    this.courseTitle = this.courseTitle.trim();
```
jrjohnson: Now that the validator accounts for this is it needed here? If so maybe it should be on a setter for courseTitle instead?

stopfstedt: this is just to trim force input cleanup on submission, otherwise leading/trailing whitespace will remain in the form component until refresh (the backend will strip it out regardless).

feels weird/redundant to set it there, but couldn't think of a better place at that moment.

PR #1042, opened 2019-12-13
Add failing tests for user-search

jrjohnson:
Exposes an issue where sending a promise for currentlyActiveUsers
doesn't cause the property to update correctly.

PR #1049, opened 2019-12-16
Fixes user search with promises

stopfstedt:
supersedes #1042


jrjohnson:
@stopfstedt changes back to you with my suggestion, not sure if I'm missing an edge case you were accounting for originally that I just didn't see.

stopfstedt:
> @stopfstedt changes back to you with my suggestion, not sure if I'm missing an edge case you were accounting for originally that I just didn't see.

no, your is much cleaner. i think i ended up with this convoluted fix due to the fact that i wasn't sure what the interface contract was exactly, e.g. trying to accommodate for a non-issue (unresolved promises as input).

PR #1050, opened 2019-12-16
Fix authentication type computed property

jrjohnson:
Was calling the wrong method to get its value. Added some tests to catch
future issues.


stopfstedt:
i feel stupid not catching this the first time around. _worst code reviewer, ever._

PR #1065, opened 2019-12-19
glimmerizes SortableTh and SortableHeader component.

stopfstedt:


jrjohnson:
I'm not sure if this points to needing a better pattern for default values, or if this case is an exception not likely to be too common.

PR #1067, opened 2019-12-20
Glimmerizes MyMaterials component

stopfstedt:


addon/components/my-materials.js
```diff
@@ -1,61 +1,76 @@
...
+      what += ':desc';
     }
-  },
+    this.args.setSortBy(what);
+  }
+
+  @restartableTask
+  *load(element, [materials]) {
+    this.loading = true;
+    this.materials = yield materials;
```
jrjohnson: Since materials are passed in as a resolved promise `@materials={{await this.materials}}` this `yield` seems unnecessary.

stopfstedt: done. simplified.

PR #1076, opened 2019-12-23
glimmerizes SingleEvent component.

stopfstedt:


addon/components/single-event.js
```diff
@@ -1,42 +1,50 @@
...
 import moment from 'moment';
 
-export default Component.extend({
-  currentUser: service(),
-  intl: service(),
-  router: service(),
-  store: service(),
-
-  classNames: ['single-event'],
+const TypedLearningMaterial = ObjectProxy.extend({
```
jrjohnson: Not sure I understand why this is an ObjectProxy and not just a class with a constructor that sets type when passed an LM.

stopfstedt: that would work, too. proxying the original LM object makes sense to me here (vs copying attributes from one object to the next), and `ProxyObject` is what Ember gives out out of the box. 
That being said, i don't feel super strong about this, so will take another stab at this.

Side-note: native JS proxies are available in all modern browsers that we're officially supporting (IE doesn't count anymore), will give that a try https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

jrjohnson: I probably don't care either, I think I wasn't sure if the original properties were even being accessed. I'm happy to try out native proxies though. Will eagerly await that experiment.

stopfstedt: ok, i replaced the ObjectProxy with a JS-native Proxy equivalent. lmk what you think.

jrjohnson: New stuff! I have no idea if I like it or not. I think go for it and let's see how it holds up in a few months when someone looks at it again.

PR #1094, opened 2020-01-08
Octanify <DetailLearningMaterials> component

jrjohnson:


addon/utils/sortable-by-position.js
```diff
@@ -0,0 +1,25 @@
+/**
```
stopfstedt: i :heart: this refactor.

jrjohnson: Me too, for realzieees. I was playing around with decorators when I realized it didn't need to be that hard!

PR #1110, opened 2020-01-14
Octaneify MeSH

jrjohnson:
Had to do both <DetailMesh> and <MeshManager> at the same time as they
were too interrelated not to.

### DIFF
--- a/addon/components/mesh-manager.js
+++ b/addon/components/mesh-manager.js
@@ -25,10 +25,10 @@ export default class MeshManagerComponent extends Component {
     this.searching = true;
     this.query = query;
 
-    const descriptors = yield this.store.query('mesh-descriptor', {
+    const descriptors = (yield this.store.query('mesh-descriptor', {
       q: query,
       limit: SEARCH_RESULTS_PER_PAGE + 1
-    });
+    })).toArray();
 
     this.searching = false;
     this.searchPage = 1;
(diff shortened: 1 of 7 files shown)


addon/components/mesh-manager.js
```diff
@@ -1,120 +1,80 @@
...
-      return ProxiedDescriptors.create({
-        content: descriptor,
-        terms: terms
-      });
+
+    this.searching = false;
+    this.searchPage = 1;
+    this.hasMoreSearchResults = (descriptors.length > SEARCH_RESULTS_PER_PAGE);
+    if (this.hasMoreSearchResults) {
+      descriptors.pop();
```
stopfstedt: this blows up. descriptors is still and `DS.PromiseArray` object at this point, which subclasses `Ember.Array`, neither implement `pop()` . Please convert `descriptors` to a native JS array further up, using `toArray()`.

addon/components/mesh-manager.js
```diff
@@ -1,120 +1,80 @@
...
+      offset: this.searchPage * SEARCH_RESULTS_PER_PAGE
     });
-    this.set('searchPage', this.get('searchPage') + 1);
-    this.set('hasMoreSearchResults', (results.length > this.get('searchResultsPerPage')));
-    if (this.get('hasMoreSearchResults')) {
-      results.pop();
+    this.searchPage = this.searchPage + 1;
+    this.hasMoreSearchResults = (descriptors.length > SEARCH_RESULTS_PER_PAGE);
+    if (this.hasMoreSearchResults) {
+      descriptors.pop();
```
stopfstedt: same thing, see comment about `pop()` above.

addon/components/mesh-manager.js
jrjohnson: Fixed, added tests even!

addon/components/mesh-manager.js
jrjohnson: Thanks for catching this, I thought me search for `Bees` was returning more than 50 results. Wrong!

PR #1122, opened 2020-01-21
Remove nullish coalescing plugin

jrjohnson:
This is provided by the default babel config now as it has moved to
stage 4.


dartajax:
Ciabatta 1.0 - cleaning up the sandwich keeping only the correct ingredients.

PR #1136, opened 2020-01-27
glimmerize SessionOverview component.

stopfstedt:
refactors publication menu out into its own subcomponent.
adds Gte (Greater Than or Equal) validator.

### DIFF
--- a/addon/components/session-overview.hbs
+++ b/addon/components/session-overview.hbs
@@ -1,384 +1,371 @@
-{{! template-lint-disable attribute-indentation }}
-<div class="session-header">
-  <span class="title">
-    {{#if @editable}}
-      {{#editable-field
-        value=this.title
-        save=(action "changeTitle")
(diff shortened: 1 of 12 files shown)


addon/components/session-overview.hbs
```diff
@@ -1,384 +1,371 @@
...
-      {{/if}}
-    </div>
-  </div>
-
-  <div class="session-overview-content">
-    <div class="sessiontype block">
-      <label>{{t "general.sessionType"}}:</label>
-      <span>
+      </span>
+      <span class="course-publication">
```
jrjohnson: Copy paste? Doesn't do anything, maybe should do something like [this](https://github.com/ilios/common/blob/master/app/styles/ilios-common/components/course-header.scss#L38)?

stopfstedt: c/p indeed. good catch! fixed.

PR #1142, opened 2020-01-29
Keep track of ILM status in component

jrjohnson:
In Ember 3.16 the computed property in the session model isn't getting
re-computed when an ILM is added / removed from the session. Instead of
delegating that to the model we can just track it in session-overview
directly.


stopfstedt:
rebased against `master` in order to trigger rebuild.

PR #1145, opened 2020-01-30
Fix detail filter on calendar

jrjohnson:
When the page was loaded directly with the detail filter open the
cohortProxies property was not yet set. Since the template looks for
this.cohorts to be an array adding an option chain here returns
undefined until the cohortProxies is set and then returns the array as
intended.

Fixes ilios/frontend#5098

### DIFF
--- a/addon/components/dashboard-calendar.js
+++ b/addon/components/dashboard-calendar.js
@@ -174,7 +174,7 @@ export default class DashboardCalendarComponent extends Component {
   }
 
   get cohorts() {
-    return this.cohortProxies.mapBy('cohort');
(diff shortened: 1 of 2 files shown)


addon/components/dashboard-calendar.js
```diff
@@ -174,7 +174,7 @@ export default class DashboardCalendarComponent extends Component {
   }
 
   get cohorts() {
-    return this.cohortProxies.mapBy('cohort');
+    return this.cohortProxies?.mapBy('cohort');
```
stopfstedt: TIL about option chaining. :+1:

PR #1152, opened 2020-02-01
Monthly Calendar as a CSS Grid

jrjohnson:
Going to a grid allows for a significant simplification of the structure of this data making is more accessible, testable, and removing the need for the elemental monthly calendar.


addon/components/monthly-calendar.js
```diff
@@ -0,0 +1,78 @@
...
+      day.dayOfWeek = date.weekday() + 1;
+      day.weekOfMonth = Math.ceil((date.date() + offset) / 7);
+      day.name = date.format('dddd LL');
+
+      return day;
+    });
+  }
+
+  get dayNames() {
+    //access the locale info here so the getter will recompute when it changes
```
stopfstedt: :heart: !!

app/styles/ilios-common/components/monthly-calendar.scss
```diff
@@ -0,0 +1,100 @@
...
+    @include ilios-heading-h4;
+    margin: 0 0 .25rem 0;
+    padding: 0;
+  }
+
+  .calendar {
+    display: grid;
+    grid-template-columns: repeat(7, 1fr);
+    grid-gap: 5px;
+    grid-template-rows: 1rem repeat(5, 7rem);
```
stopfstedt: February 2027 will have a blank row, since all of its days fit within four weeks. I think this edge case is acceptable.

Selection_024

PR #1155, opened 2020-02-03
Octanify Course Objectives

jrjohnson:
Stopped relying on the shared <DetailObjectives> component and instead moved all of this into a brand new set of course specific objective components. This allowed by to remove a very old and buggy pattern where changes for MeSH and parents were made on the objective directly instead of a buffer as well as simplify the entire thing since it no longer needed to account for being used with a session or program year.

I swear this was the smallest possible changeset to make that work 😝 

It does make the course objective manager for choosing parents a **lot** faster, so that's nice!

Bugs to fix:
- [x] adding a new objective to empty course fails
- [x] if course has no cohorts and some are added they are not immediately available for picking parent objectives, a refresh is necessary.
- [x] Test: `Acceptance | Course - Competencies`


addon/components/course/objectives.hbs
```diff
@@ -0,0 +1,103 @@
...
+          type="button"
+          class="bigcancel"
+          {{on "click" this.cancel}}
+          aria-label={{t "general.cancel"}}
+          data-test-cancel
+        >
+          <FaIcon @icon="undo" />
+        </button>
+      {{else if @editable}}
+        <LinkTo @route="course-visualize-objectives" @model={{@course}}>
```
stopfstedt: even if the course objectives are not editable by the current user - shouldn't they still be able to use the visualizations?

addon/components/course/objectives.js
```diff
@@ -0,0 +1,215 @@
...
+        title: `${program.title} ${cohort.title}`,
+        id: cohort.id,
+        competencies,
+      };
+    });
+  }
+
+  @dropTask
+  *manageParents(objective) {
+    const cohortObjectives = yield waitForProperty(this, 'cohortObjectives', Array.isArray);
```
stopfstedt: this looks like it has the potential to lock up.
just thinking out loud here - would it be possible to init `cohortObjectives` as an empty array, and then check here (and elsewhere, template-side) if cohortObjectives not empty before allowing parent management?

addon/components/course/objectives.hbs
jrjohnson: They aren't right now https://github.com/ilios/common/blob/master/addon/components/detail-objectives.hbs#L37

addon/components/course/objectives.js
jrjohnson: Yea, I have this on my list. To deactivate the button would require passing the status of this task down through several levels of component. I figured less change here is better and I'll refactor out `<CourseObjectiveList>` and '<CourseObjectiveListItem>` as a next step.

stopfstedt: fair enough.

addon/components/course/objectives.hbs
stopfstedt: i accept this.

stopfstedt:
refs ilios/frontend#4076

PR #1156, opened 2020-02-03
replace RSVP promise with async/await

stopfstedt:
geez, this code is old. 
step 0 on our path to getting OfferingForm octaneified and improved.

### DIFF
--- a/addon/components/offering-form.js
+++ b/addon/components/offering-form.js
@@ -3,12 +3,7 @@ import { computed } from '@ember/object';
 import { not } from '@ember/object/computed';
 import { inject as service } from '@ember/service';
 import { isEmpty, isPresent } from '@ember/utils';
-import {
-  Promise as RSVPPromise,
-  filter,
-  hash,
-  map
-} from 'rsvp';
+import { filter, hash, map } from 'rsvp';
 import moment from 'moment';
 import { validator, buildValidations } from 'ember-cp-validations';
 import ValidationErrorDisplay from 'ilios-common/mixins/validation-error-display';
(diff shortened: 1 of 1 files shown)


addon/components/offering-form.js
```diff
@@ -98,37 +93,29 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
...
-          }, []);
-
-          resolve(flat);
-        });
+  associatedSchools: computed('cohorts.[]', async function(){
+    const cohorts = this.get('cohorts');
+    if (isEmpty(cohorts)) {
+      return [];
+    } else {
+      const cohortSchools = map(cohorts.toArray(), cohort => {
```
jrjohnson: Should be `const cohortSchools = await map`, yes?

addon/components/offering-form.js
```diff
@@ -98,37 +93,29 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
...
+      schools.pushObjects(cohortSchools);
+      return schools.uniq().toArray();
+    }
+  }),
+  availableInstructorGroups: computed('associatedSchools.[]', async function(){
+    const associatedSchools = await this.get('associatedSchools');
+    const allInstructorGropus = await map(associatedSchools, school => {
+      return school.get('instructorGroups');
     });
+    return allInstructorGropus.reduce((flattened, obj) => {
```
jrjohnson: While you're at it this can now almost certainly be replace with the new [Array.flat()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat).

addon/components/offering-form.js
stopfstedt: thanks. fixed.

addon/components/offering-form.js
stopfstedt: the inner arrays are not real arrays, they're DS.PromiseArray objects. leaving the current code as-is, since it dumps these out into proper arrays in the process of flattening.

jrjohnson: 👍

PR #1162, opened 2020-02-05
octanifies search box. 

stopfstedt:
needed as sub-sub component of offering form (and many others).

### DIFF
--- a/addon/components/search-box.hbs
+++ b/addon/components/search-box.hbs
@@ -1,14 +1,18 @@
-<input
-  class="search-input"
-  type="search"
-  value={{this.value}}
-  placeholder={{@placeholder}}
-  {{on "input" (action "update" value="target.value")}}
->
-<span
-  class="search-icon"
-  role="button"
-  {{on "click" (pipe-action (action "focus") (perform this.searchTask))}}
->
-  <FaIcon @icon="search" />
-</span>
+<div class="search-box" data-test-search-box>
+  <input
+    class="search-input"
+    type="search"
+    value={{this.value}}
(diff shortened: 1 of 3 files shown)


addon/components/search-box.hbs
```diff
@@ -1,14 +1,18 @@
...
+    value={{this.value}}
+    placeholder={{@placeholder}}
+    {{on "input" (action "update" value="target.value")}}
+    {{on "keyup" (fn this.keyUp)}}
+    {{ref this "searchInput"}}
+  >
+  <span
+    class="search-icon"
+    role="button"
+    {{on "click" (pipe-action (fn this.moveFocus this.searchInput) (perform this.searchTask))}}
```
jrjohnson: I really like this, putting the reffed element as an argument makes the JS much nicer because there isn't a mystery property suddenly accessed.

addon/components/search-box.js
```diff
@@ -1,57 +1,58 @@
...
+  }
+
+  @action
+  keyUp({ key }) {
+    switch (key) {
+    case 'Enter':
+      this.searchTask.perform();
+      break;
+    case 'Escape':
+      if (this.args.clear) {
```
jrjohnson: Feels like this should reset the input to empty even if no `clear` method is passed. Technically an enhancement I guess, so maybe skip it.

stopfstedt: agreed. the `clear()` action already does this. invoking that one instead...

stopfstedt: ..done.

PR #1184, opened 2020-02-17
Fix boolean expression in week-glance.js

Trott:
The boolean expression checks .midnightAtTheStartOfThisWeek twice. One of
them was meant to be .midnightAtTheEndOfThisWeek instead.

Boolean OR was used where Boolean AND was meant. (If it is not midnight
at the begining of the week OR it is not midnight at the end of the
week, then the expression will always be TRUE. If AND is used instead, it excludes
midnight at the start or end of the week.)

### DIFF
--- a/addon/components/week-glance.js
+++ b/addon/components/week-glance.js
@@ -35,7 +35,7 @@ export default class WeeklyGlance extends Component {
   }
 
   get title() {
-    if (!this.midnightAtTheStartOfThisWeek || !this.midnightAtTheStartOfThisWeek) {
+    if (!this.midnightAtTheStartOfThisWeek && !this.midnightAtTheEndOfThisWeek) {
       return '';
     }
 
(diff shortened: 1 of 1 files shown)


Trott:
I have no idea if/how this code is exercised in the tests, so I did not write a test. Assistance welcome.

stopfstedt:
> The boolean expression checks .midnightAtTheStartOfThisWeek twice. One of
> them was meant to be .midnightAtTheEndOfThisWeek instead.
correct.
> 
> Boolean OR was used where Boolean AND was meant. (If it is not midnight
> at the begining of the week OR it is not midnight at the end of the
> week, then the expression will always be TRUE. If AND is used instead, it excludes
> midnight at the start or end of the week.)
not correct. this is just a safeguard against those attributes not being set yet. if either of them are falsy, it should return early with a blank string.

PR #1185, opened 2020-02-17
Fix boolean expression in not-blank validation decorator

Trott:
Boolean OR was used where AND was intended. As previously written, the
expression will always evaluate to TRUE.


Trott:
As with the other one, I'm not sure how to write a test for this because I'm unfamiliar with Ilios tests overall, but this seemed like a pretty clear error and local tests pass with the change.

stopfstedt:
i see the fallacy here, thanks for the fix.

should probably always return `FALSE` if the input is _not_ a string.

PR #1189, opened 2020-02-20
alternative component validator

stopfstedt:
hybrid approach to component validation by using a class decorator for plumbing (validation registry etc), and using class getters for porcelain (the specific field validation implementations).

the getters are a bit long-handed atm, but can be refactored for improved usability.

this should regain us the flexibility to build arbitrarily complex component validations. 

todo:

- [ ] decorator test coverage
- [ ] a second component conversion - use `course-rollover`.

other things of note:
- unlike the original decorators that were built on top of the `class-decorator` library, this implementation is not asyncronous. this can be changed if necessary if need be down the road.

### DIFF
--- a/addon/components/course-overview.hbs
+++ b/addon/components/course-overview.hbs
@@ -65,10 +65,10 @@
             <Input
               type="text"
               @value={{this.externalId}}
-              @key-press={{fn this.addErrorDisplayFor "externalId"}}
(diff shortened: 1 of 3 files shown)


stopfstedt:
i'm scrapping this.
- can't do async, which we need down the road (unique user name validations in user mgmt. components)
- error display state management is awkward
- the current event bindings do not jive with ember-cli-page-objects

back to class-validator and custom decorators :(((((

PR #1193, opened 2020-02-22
[breaking] Smell of Octane in the Morning

jrjohnson:
All of the `<Leadership*>`  components now require passing the `showDirectors/showAdministrators` options. Since they had originally defaulted to true it's not possible to use a pattern like `{{if (or @showDirectors true)}}` as it will always evaluate to true.

Mostly this is contained to within common, but there are used in the frontend in a few places.


addon/components/editable-field.js
```diff
@@ -1,54 +1,49 @@
...
-  }).drop(),
+    yield this.args.close();
+    this.isEditing = false;
+  }
 
-  edit: task(function * () {
-    this.set('isEditing', true);
-    yield timeout(10);
-    const elements = this.element.querySelectorAll('input,textarea,select,.fr-element');
+  focusFirstControl(element) {
```
stopfstedt: :heart:

PR #1207, opened 2020-03-02
[breaking] My Oh My Octane!

jrjohnson:
Breaking removal of `ember-one-way-select` will need to be re-added to the frontend.

### DIFF
--- a/addon/decorators/validation/after-date.js
+++ b/addon/decorators/validation/after-date.js
@@ -8,7 +8,7 @@ export function AfterDate(property, validationOptions) {
       name: "afterDate",
       target: object.constructor,
       propertyName: propertyName,
-      constraints: [property, validationOptions ?? 'second'],
+      constraints: [property],
       options: validationOptions,
       validator: {
         validate(value, { constraints, object: target, property }) {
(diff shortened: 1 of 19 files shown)


addon/decorators/validation/after-date.js
```diff
@@ -8,11 +8,11 @@ export function AfterDate(property, validationOptions) {
       name: "afterDate",
       target: object.constructor,
       propertyName: propertyName,
-      constraints: [property],
+      constraints: [property, validationOptions ?? 'second'],
```
stopfstedt: Looking into the class validator [docs](https://github.com/typestack/class-validator#custom-validation-decorators), it  
appears that `validationOptions` is expected to be an object.

I'm concerned that passing a string may clobbering something internally.

Perhaps change it to have it invoked in this form `@BeforeDate('endDate', {granularity: 'day'})`.  That would also leave us the option to pass other arguments, like the `message` attribute that we've discussed.

jrjohnson: Great catch, fixed! Also enjoy my use of both optional chaining and null coalescing at the same time!!

stopfstedt: sweet! looks good.

PR #1215, opened 2020-03-02
octanify offering form

stopfstedt:


addon/components/offering-form.js
```diff
@@ -129,23 +67,25 @@ export default Component.extend(ValidationErrorDisplay, Validations, {
...
+    return mEnd.diff(mStart, 'hours');
+  }
 
-    return diffInHours;
-  }),
-  durationMinutes: computed('startDate', 'endDate', function(){
-    const startDate = this.get('startDate');
-    const endDate = this.get('endDate');
+  @IsInt() @Gte(0) @Lte(59)
+  get durationMinutes() {
```
jrjohnson: This is fun! Sort of seems like maybe the future.

addon/decorators/validation/gt.js
```diff
@@ -0,0 +1,38 @@
...
+      propertyName: propertyName,
+      constraints: [gt],
+      options: validationOptions,
+      validator: {
+        validate(value, { constraints, property }) {
+          if (!constraints || constraints.length < 1) {
+            throw new Error(`You must pass a ${property} value as the first argument to Gte`);
+          }
+          const numValue = Number(value);
+          if (typeof numValue !== 'number' || isNaN(numValue)) {
```
jrjohnson: It doesn't look like `Number()` can ever return anything but a number or `NaN` so the `typeof` comparison seems superfluous.

addon/decorators/validation/is-int.js
```diff
@@ -0,0 +1,32 @@
...
+      target: object.constructor,
+      propertyName: propertyName,
+      options: validationOptions,
+      validator: {
+        validate(value) {
+          const numValue = Number(value);
+          if (Number.isInteger) {
+            return Number.isInteger(numValue);
+          }
+          // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger#Polyfill
```
jrjohnson: It looks like this exists in all of our supported browsers, is there a specific one that I missing that would need a polyfill for this?

addon/decorators/validation/gt.js
stopfstedt: you're right, good catch. yanked from all existing number comparison validators.

addon/decorators/validation/is-int.js
stopfstedt: internet exploder doesn't support this method. but even tho we don't support IE anymore, it costs us virtually nothing to polyfill this method. not feeling super strong about this, so if you want this gone then i'll yank it.

addon/components/offering-form.js
stopfstedt: yea, chaining them is the way to go.

addon/decorators/validation/is-int.js
jrjohnson: It's fine with me, you're right no big deal.

jrjohnson:
@dartajax this needs lots of click testing in all the places where an offering can be added or modified. If it passes, it's good to merge.

PR #1220, opened 2020-03-05
fixes scroll into view upon offering form closing.

stopfstedt:
this is an un-ticketed regression that i found when working on something else.

in the sessions list, when you close the embedded offering form, you end up with this dreadful error message:

```
Uncaught Error: You attempted to access the 'element' property on a Glimmer Component, but that property does not exist in Ember.js applications, it only exists in Glimmer.js apps. You define a class field with the same name on your component class and it will overwrite this error message, but it will not be used by the framework.
```

this was an oversight during glimmerization of this component, using a ref-modifier fixes that issue.


jrjohnson:
Rebase should fix tests.

PR #1222, opened 2020-03-06
corrects passing of title attribute to the icon.

stopfstedt:
just saw that in passing, no big deal...

four eyes see better than two


jrjohnson:
Rebase should fix tests.

stopfstedt:
in addition to persisting dependency conflicts, this now has actual integration test failures. will update the test coverage and resubmit this for review.

PR #1242, opened 2020-03-16
[breaking] Here's looking at you Octane

jrjohnson:
More components on the Octane Train.  Breaking change as `<PublicationStatus>` now requires that the `showText` and `showIcon` values be passed in.

### DIFF
--- a/addon/components/print-course.hbs
+++ b/addon/components/print-course.hbs
@@ -1,331 +1,117 @@
-<div class="header" data-test-course-header>
-  <h2>
-    {{@course.title}}
-  </h2>
-  <h4>
-    {{@course.academicYear}}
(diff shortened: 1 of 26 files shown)


addon/components/print-course.hbs
```diff
@@ -1,331 +1,117 @@
...
+      <br>
       <div class="inline-label-data-block">
         <label>
-          {{t "general.description"}}:
+          {{t "general.directors"}}:
         </label>
         <div>
-          {{session.sessionDescription.textDescription}}
+          <span>
+            {{#each (sort-by "lastName" "firstName" (await @course.directors)) as |user|}}
```
stopfstedt: nice!

PR #1245, opened 2020-03-17
scroll bottom of offering form into view when it gets opened up.

### LINKED ISSUE #899: Clicking to Add New Offering Should Navigate User Further Down the Screen

stopfstedt:
fixes #899

### DIFF
--- a/addon/components/offering-form.hbs
+++ b/addon/components/offering-form.hbs
@@ -1,8 +1,8 @@
 <div
   class="offering-form"
   data-test-offering-form
-  {{did-insert (perform this.load) @offering @cohorts}}
-  {{did-update (perform this.load) @offering @cohorts}}
+  {{did-insert (perform this.load) @offering @cohorts @scrollToBottom}}
+  {{did-update (perform this.load) @offering @cohorts @scrollToBottom}}
 >
   {{#unless this.loadData.isRunning}}
     <div class="toggle-offering-calendar">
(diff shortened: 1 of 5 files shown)


addon/components/offering-form.hbs
```diff
@@ -244,7 +244,7 @@
           </p>
         {{/each}}
       </div>
-      <div class="buttons">
+      <div class="buttons" {{ref this "scrollTo"}}>
```
jrjohnson: `{{ref}}` is currently bound to the ember runloop and sometimes these references aren't set when expected. I'd suggest adding a specific `{{did-insert}}` directly here that way you're sure you get the element reference and you can still check the `@scrollToBottom` value.

stopfstedt: i don't see a way to chain `did-insert` modifiers. and i don't think they stack in order specified. or do they?

most of the markup in this  component gets rendered after the `loadData` task has completed, including the buttons container that we want to scroll to. so whatever code invokes scroll-into-view will have to run afterwards.
that's why i'm running this sequentially within the `load` task.

PR #1252, opened 2020-03-19
Octane objective lists

jrjohnson:
Replaced both course and session lists with improved specific components that don't rely on a mixin to do their job.


addon/components/course/objective-list.hbs
```diff
@@ -1,23 +1,27 @@
 <div
   class="course-objective-list"
-  {{did-insert (perform this.load) (await @course.objectives)}}
-  {{did-update (perform this.load) (await @course.objectives)}}
   data-test-course-objective-list
+  {{did-insert (perform this.load) @course @course.objectives}}
```
stopfstedt: the `load()` task receives a second parameter with `@course.objectives` here, but it is not being processed. kruft or deliberate decision to pass this?

addon/components/session/objective-list.hbs
```diff
@@ -0,0 +1,100 @@
+<div
+  class="session-objective-list"
+  data-test-session-objective-list
+  {{did-insert (perform this.load) @session @session.objectives}}
```
stopfstedt: same question as above. is the second argument passed in order to force a re-render?

addon/components/course/objective-list.hbs
jrjohnson: It isn't used in `load` but load should be called again when it changed. It probably only needs to go on `did-update` to trigger the change, but I left it in both places because otherwise it looks like it is missing from one.

addon/components/session/objective-list.hbs
jrjohnson: Yup. Exactly that. It's the thing that is being tracked, but the real state of it is a promise so I didn't want to await it here. Instead just let ember do it's thing which will mark this property as needing to be updated and call `load` again.

dartajax:
LGTM as well.

PR #1262, opened 2020-03-23
link learners to offerings.

### LINKED ISSUE #909: Update Offering Manager to allow for association of individual learners to offerings, as well as learner groups
We need to refactor the offering manager to include the ability to select and add individual learners to an event, along with the learner group association tools.

stopfstedt:
Selection_089

fixes #909

### DIFF
--- a/addon/components/detail-learner-list.hbs
+++ b/addon/components/detail-learner-list.hbs
@@ -1,0 +1,12 @@
+<ul class="detail-learner-list" data-test-detail-learner-list>
+  {{#each (sort-by "fullName" (await @learners)) as |user|}}
+    {{#if @isManaging}}
+      <li role="button" {{on "click" (fn @remove user)}}>
+        {{user.fullName}}
+        <FaIcon @icon="times" class="remove" />
+      </li>
+    {{else}}
+      <li>{{user.fullname}}</li>
+    {{/if}}
+  {{/each}}
+</ul>
--- a/addon/components/detail-learner-list.hbs
+++ b/addon/components/detail-learner-list.hbs
@@ -2,11 +2,11 @@
   {{#each (sort-by "fullName" (await @learners)) as |user|}}
     {{#if @isManaging}}
       <li role="button" {{on "click" (fn @remove user)}}>
-        {{user.fullName}}
+        <span data-test-user-name>{{user.fullName}}</span>
         <FaIcon @icon="times" class="remove" />
       </li>
     {{else}}
-      <li>{{user.fullname}}</li>
+      <li><span data-test-user-name>{{user.fullName}}</span></li>
     {{/if}}
   {{/each}}
 </ul>
--- a/addon/components/detail-learner-list.hbs
+++ b/addon/components/detail-learner-list.hbs
@@ -1,12 +1,8 @@
 <ul class="detail-learner-list" data-test-detail-learner-list>
   {{#each (sort-by "fullName" (await @learners)) as |user|}}
-    {{#if @isManaging}}
-      <li role="button" {{on "click" (fn @remove user)}}>
-        <span data-test-user-name>{{user.fullName}}</span>
-        <FaIcon @icon="times" class="remove" />
-      </li>
-    {{else}}
-      <li><span data-test-user-name>{{user.fullName}}</span></li>
-    {{/if}}
+    <li role="button" {{on "click" (fn @remove user)}}>
+      <span data-test-user-name>{{user.fullName}}</span>
+      <FaIcon @icon="times" class="remove" />
+    </li>
   {{/each}}
 </ul>
--- a/addon/components/offering-form.hbs
+++ b/addon/components/offering-form.hbs
@@ -222,16 +222,18 @@
           />
         </div>
       {{/if}}
-      <div class="learners item">
-        <label>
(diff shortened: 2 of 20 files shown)


saschaben:
testing feedback:
1) on initial creation of an offering, need to change form validation to have either individual learner, or learner group required. currently learner group is required -- so if you add no group but five users, validation fires and you cannot save. This appears to NOT be the case for editing an existing offering (you can remove all groups and leave only individual learners, and save).
2) we will need to provide for display of selected learners on offerings, not just groups (as we do now)
3) generally -- it appears to work! woo hoo!

stopfstedt:
> testing feedback:
> 
> 1. on initial creation of an offering, need to change form validation to have either individual learner, or learner group required. currently learner group is required -- so if you add no group but five users, validation fires and you cannot save. This appears to NOT be the case for editing an existing offering (you can remove all groups and leave only individual learners, and save).
> 2. we will need to provide for display of selected learners on offerings, not just groups (as we do now)
> 3. generally -- it appears to work! woo hoo!

thanks for the quick feedback. 
1. yes, working on it.
2. agreed, but should be addressed as separate issue. is this a blocker?
3. w00t!

addon/components/detail-learner-list.hbs
```diff
@@ -0,0 +1,12 @@
+<ul class="detail-learner-list" data-test-detail-learner-list>
+  {{#each (sort-by "fullName" (await @learners)) as |user|}}
+    {{#if @isManaging}}
+      <li role="button" {{on "click" (fn @remove user)}}>
```
jrjohnson: Can this just be a button inside of the LI? Instead of the role?

dartajax:
While it is possible to add a learner who is already a member of an attached learner group, I don't see that as an immediate problem. Maybe we should in the future grey out learners who are already attached to the offering as a member of a learner group.

Nice work. It does indeed work. @saschaben's comments are ones that I agree with.

addon/components/offering-form.js
```diff
@@ -146,6 +147,16 @@ export default class OfferingForm extends Component {
...
   }
 
+  @action
+  async addLearner(learner) {
+    this.learners = [...this.learners, learner];
+  }
+
+  @action
+  async removeLearner(learner) {
+    this.learners = this.learners.filter(l => l !== learner);
```
jrjohnson: I've recently seen some performance pattern with comparing ember-data models. They're pretty deep objects with a lot of metadata. Probably better to check the ID if possible.

stopfstedt:
@saschaben @dartajax on further inspection - learner groups are optional thourought, _except_ when creating small group offerings. updating existing offerings, or creating a single offering does not require a learnergroup to be attached.
correct me if i'm wrong, but learners are not mandatory when creating small groups. therefore, the existing lack of validation on this attribute seems to be correct to me.
thoughts?

addon/components/detail-learner-list.hbs
stopfstedt: i agree on the principle, but in practice this would (and should) be a larger undertaking. we should look and address this in all of those tag-list-style components.

Selection_097

addon/components/offering-form.js
stopfstedt: good point, fixed.

dartajax:
It seems to me it is possible to create an `offering` without assigning the learner group. It also seems that it is not possible to create a `small group` offering without attaching a learner group.

dartajax:
Ah but you can later remove the learner group from a `small group`offering.

saschaben:
Correct on all counts above. But right now the small group generator by definition needs some groups to instantiate -- maybe we can create an option for "number of small groups" when no groups are selected. but even more importantly: when creating small groups, _how do we assign individual learners to the instantiated small group offerings?_

stopfstedt:
> Correct on all counts above. But right now the small group generator by definition needs some groups to instantiate -- maybe we can create an option for "number of small groups" when no groups are selected. but even more importantly: when creating small groups, _how do we assign individual learners to the instantiated small group offerings?_

As currently implemented, all individually selected users will become part of the each created small group offering. 

example: in small-group mode, i create a new offering with learners A and B, and learner groups X and Y attached. this spawns off two new offerings, one X, A and B attached, and one with Y, A and B attached.

seems weird. perhaps suppress the learner attachment control in new small-group offering mode of this form?

saschaben:
> perhaps suppress the learner attachment control in new small-group offering mode of this form?

Until we think it thru further, this might be the most appropriate way to go. we always run into trouble when we try to auto-assign anyone to anything. so: suppress on small group `creation`; allow on single session creation; and hopefully allow on `edit` for any created offering...

addon/components/offering-form.hbs
```diff
@@ -222,6 +222,18 @@
...
         </div>
       {{/if}}
+      {{#unless @smallGroupMode}}
+        <div class="learners item">
+          <label>
+            {{t "general.learners"}}:
+          </label>
+            <LearnerSelectionManager
+              @learners={{this.learners}}
+              @add={{fn this.addLearner}}
```
jrjohnson: `fn` is unneeded here since `addLearners` is a function already and there are no additional arguments to be added at this point.

stopfstedt: fixed.

PR #1263, opened 2020-03-23
Pass IDs for Dashboard Filters

jrjohnson:
Passing the model was causing some performance issues, especially in comparing and displaying the active filters. Instead this passes only the IDs.


addon/components/toggle-buttons.js
```diff
@@ -13,7 +13,6 @@ export default class ToggleButtons extends Component {
     if (! this.args.firstOptionSelected) {
       this.args.toggle(true);
     }
-    console.log(this.elementId);
```
stopfstedt: wheeee!

jrjohnson: 😆 forgot that was in there. I was super confused why my console was spitting out these weird `undefined` messages. Probably need to add an ESLint rule for that!

PR #1266, opened 2020-03-24
Fix checkbox style in PublishAllSessions component

### LINKED ISSUE #1265: Broken Styles on Publish All Sessions
Looking bad!

Screen Shot 2020-03-23 at 11 15 09 PM

jrjohnson:
Instead of using the span as a button this just uses the HTML label to
do the same thing.

Fixes #1265


dartajax:
LGTM 👍

PR #1267, opened 2020-03-25
Replace mixins with controllers

jrjohnson:
Ember seems to handle these correctly from an addon now and mixins are an anti-pattern.


dartajax:
I did a click test as well. Gonna merge.

PR #1269, opened 2020-03-25
Use tracked for query params

jrjohnson:
Switch back to `@tracked` to be more Octane Idiomatic.

Requires https://github.com/emberjs/ember.js/issues/18715 to be fixed. This is a reproduction for checking in on that issue.


stopfstedt (approved):
LGTM

PR #1277, opened 2020-03-29
Improved year based filters

jrjohnson:
Updated both the course and cohort filters to contain or not need the year anymore.

Fixes ilios/frontend#1830

### DIFF
--- a/addon-test-support/ilios-common/page-objects/components/dashboard/cohort-calendar-filter.js
+++ b/addon-test-support/ilios-common/page-objects/components/dashboard/cohort-calendar-filter.js
@@ -1,0 +1,19 @@
+import {
+  collection,
+  clickable,
+  create,
+  property,
+  text,
+} from 'ember-cli-page-object';
+
+const definition = {
+  scope: '[data-test-cohort-calendar-filter]',
+  cohorts: collection('[data-test-cohort]', {
+    title: text(),
+    toggle: clickable('[data-test-target]'),
+    isChecked: property('checked', '[data-test-target] input'),
+  }),
+};
+
+export default definition;
+export const component = create(definition);
--- a/addon-test-support/ilios-common/page-objects/components/dashboard/courses-calendar-filter.js
+++ b/addon-test-support/ilios-common/page-objects/components/dashboard/courses-calendar-filter.js
@@ -1,0 +1,23 @@
+import {
+  collection,
+  clickable,
+  create,
+  property,
+  text,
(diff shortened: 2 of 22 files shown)


saschaben:
Comments: With the new presentation of courses, it makes it even more desirable to be able to change the filter labels ("Courses", "Program/Cohort", etc) slightly. Since they scroll out of view as part of the box content now, we might want to consider either: moving them out of the box to be headers above the filters, OR (if that is more effort than is reasonable for the result), modifying the label of the course years from  "2019-2020 >" etc. to "2019 - 2020 Courses >" just to provide context, especially since we now open at current year (which will make any older school lose the current label to the scrolling).

Otherwise, I think this is a good step forward.

jrjohnson:
Thanks @saschaben I forgot to take care of that. I've locked the headers for each filter in place and changed the styling a bit. let me know what you think.

saschaben:
Now that's rock'n'roll. we will likely want to dial in the headers visually a bit (vertical alignments seem crowded right now) but that can be addressed later. I approve of this message. let's get it in front of users for review.

dartajax:
Here's what I suggest. It's going to cause brain pain. I apologize for that and will not be hurt if this gets vetoed.

Regarding the default year expanded in the new Courses filter, I am going to propose the following ...

Default year expanded
2018-2019 when the current date is between April 1, 2018 and April 1, 2019
2019 2020 when the current date is between April 1, 2019 and April 1, 2020
2020-2021 when the current date is between April 1, 2020 and April 1, 2021.

Why this cockamamie date scheme? I understand I am making assumptions here and we all know what can happen with those. Course coordinators and admins start wanting to see the following year's courses in Spring usually to see what's coming up. I chose April 1st - maybe May 1st would be better. If this is even considered, I would be glad to find the best cutoff date by asking users. 

In lieu of this, we could just use July 1 - June 30th as the guideline. That would be fine too.

dartajax:
In the screen shot below, there is no way of knowing which year these courses pertain to. 

image

Is there any way to make it look more like this even if the user has scrolled down a bit?

image

It might not be technically feasible since the same thing could apply to vocabularies where DEI, Resource Types, etc. are similar and could remain displayed so the user knows what they are accessing.

jrjohnson:
@dartajax give this another look. I've changed the expanded year to your suggested April dates and when the year title goes off the screen I've added it to the header. I think some formatting might be needed to make it clear why it is up there, but I think this is better than having multiple scroll boxes in the same space.

PR #1293, opened 2020-04-02
Remove Publishable Model Mixin

jrjohnson:
A step on the way to turning models into ES classes. Gotta get rid of
them mixins.


dartajax:
LGTM 👍

PR #1300, opened 2020-04-06
[breaking] Replace mixin routes with ES Classes

jrjohnson:
We no longer need mixins to support routes in both the dummy app and the
frontend. I've also converted the routes to modern the modern octane
patterns and brought in the templates from the frontend.


jrjohnson:
I was so surprised it worked! I think the hangup in the past was the dashboard route. I'm looking into how to include that as well, but need to move my-reports over. Got about half way there but figured smaller PR and then another PR was a better strategy 😁

dartajax:
I clicked around too. Gonna approve and merge this sapsucker.

PR #1335, opened 2020-04-17
Cache the permission matrix

jrjohnson:
When a user without high level permissions visits a course the matrix
can be loaded many times while checking permissions. This causes the
schools to be fetched each time and the array re-built. Instead we can
cache this mostly static value.

### DIFF
--- a/addon/services/permission-matrix.js
+++ b/addon/services/permission-matrix.js
@@ -4,13 +4,19 @@ import { deprecate } from '@ember/debug';
 
 export default Service.extend({
   store: service(),
+  _permissionMatrixPromise: null,
   permissionMatrix: computed(async function(){
     deprecate('Async Computed Called', false, {id: 'common.async-computed', until: '40'});
     return this.getPermissionMatrix();
   }),
   async getPermissionMatrix() {
-    const store = this.get('store');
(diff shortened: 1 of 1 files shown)


addon/services/permission-matrix.js
```diff
@@ -4,13 +4,19 @@ import { deprecate } from '@ember/debug';
...
+  _permissionMatrixPromise: null,
   permissionMatrix: computed(async function(){
     deprecate('Async Computed Called', false, {id: 'common.async-computed', until: '40'});
     return this.getPermissionMatrix();
   }),
   async getPermissionMatrix() {
-    const store = this.get('store');
-    const schools = await store.findAll('school');
+    if (!this._permissionMatrixPromise) {
+      this._permissionMatrixPromise = this._fillMatrix();
```
stopfstedt, line 14: shouldn't the `await` be on this line?

jrjohnson: No because it's storing the promise itself and not the result. This keeps it non-blocking so the next call doesn't also start a request, it just moves on and waits for the one that is in flight already.

stopfstedt: thanks for clarifying.

PR #1340, opened 2020-04-20
Octaniefy <SessionCopy> component

jrjohnson:
I removed the validations here because they would never fire. Since the
save button is disabled when course or year isn't selected it wasn't
possible to ever be in an invalid state other than when there are no
courses in the year and that is represented in the UI bay "None" courses
and a disabled save button.


stopfstedt:
side-note: once #1341 gets merged, ember-cp-validations can be yanked from common altogether.

PR #1351, opened 2020-04-22
Replace save status modal with a live button

### LINKED ISSUE #1345: Wait Saving Not Going Away on Certain Save Events
This seems to be the case when entering really long unbroken strings into the location ...

Screen shot below. I have also had this occur when removing an instructor from an offering.

image

jrjohnson:
Because WaitSaving uses a modal dialog that gets wormholed out to a
different location in the DOM it is susceptible to missing it's queue to
be destroyed when the component containing it is removed. Instead of a
modal this makes the button into their own live saving state with a
spinner, percentage, and check mark when complete.

Fixes #1345

### DIFF
--- a/addon/components/offering-form.hbs
+++ b/addon/components/offering-form.hbs
@@ -257,21 +257,18 @@
         {{/each}}
       </div>
       <div class="buttons" {{ref this "scrollTo"}}>
-        <button
+        <SaveButton
           class="done text"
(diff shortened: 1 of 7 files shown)


dartajax:
I was seeing an occasional screen flicker so this is just a note of that possibility. Other than that, I like this solution a lot.

PR #1357, opened 2020-04-22
centers text deletion dialog.

### LINKED ISSUE #1356: session and course objectives deletion dialog misalignment
Selection_127

rrrrregression! message should be centered.

stopfstedt:
fixes #1356


jrjohnson:
I put it over to the right to be near the trashcan when you press it so your eyes don't have to wander too far. That's just my preference though. I'm fine with either way, but will leave it up to @saschaben and @dartajax for a final decision.

PR #1365, opened 2020-04-29
Indicate if a session has prerequisites

### LINKED ISSUE #1364: Indicator Needed In Session List if Linked Sessions Exist
It would be helpful for course coordinators / administrators to know whether a session has pre-work linked to it.

The screen shot below shows a session that has instructional notes associated with it. It would be helpful if we could give linked sessions the same or similar treatment. If there pre-work attached, there should be an icon displayed with a similar callout message except it would designate the existence or linked pre-work rather than instructional notes.

image

jrjohnson:
Shows up on the session grid to make it clear which sessions have
prerequisites.

Fixes #1364

Wip:
- [x] Is this the right icon?


dartajax:
I love it. The icon is nice too. Good choice.

PR #1375, opened 2020-05-05
Add edit button to learning material list

### LINKED ISSUE #1374: Missing Functionality - Add MeSH to Learning Material
image

jrjohnson:
Does the same thing as clicking the LM title, but might be a little bit
more obvious for users who don't want to unintentionally download the
LM file itself.

Fixes #1374

### DIFF
--- a/addon/components/detail-learning-materials.hbs
+++ b/addon/components/detail-learning-materials.hbs
@@ -169,13 +169,12 @@
               </td>
               <td class="text-left text-center" colspan="1">
                 {{#if @editable}}
-                  <span
-                    class="clickable remove"
-                    role="button"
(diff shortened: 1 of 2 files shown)


dartajax:
Here is a screen shot from the user guide. I was going to replace this but then I noticed the "Add MeSH" button was missing completely. We must have changed this at some point. See arrow and #4 for reference.

image

dartajax:
When you are in Courses and Sessions as a Course Administrator or Coordinator, it would seem to me the title of a Learning Material would just lead to edit mode not download mode since these folks would want to perform usage maintenance on the LM and not to download it since they aren't students and probably are the people that originally attached the LM anyway. They don't need to learn about the LM as a student would. That having been said, it has been that always so the edit button makes sense.

PR #1388, opened 2020-05-10
Replace pikaday with flatpickr for dates

jrjohnson:
First step in removing our reliance on moment.js which its authors
encourage.

Todo:
- [x] feedback on new picker (usability, colors)
- [x] replace remaining date pickers
- [x] remove `ember-pikaday`


addon/components/course-rollover.hbs
```diff
@@ -68,13 +68,10 @@
...
         </label>
-        <PikadayInput
-          @size={{10}}
+        <Course::RolloverDatePicker
+          @course={{@course}}
           @value={{this.startDate}}
-          @minDate={{@course.startDate}}
-          @format="YYYY-MM-DD"
-          @onSelection={{set this.startDate}}
-          @options={{hash disableDayFn=this.disableDayFn course=@course}}
```
stopfstedt, line 77: my "beautiful" hack - gone!

addon/components/course/rollover-date-picker.js
```diff
@@ -0,0 +1,58 @@
...
+      onChange: ([selectedDate]) => {
+        // if a date is forced that isn't allowed
+        if (!selectedDate) {
+          selectedDate = course.startDate;
+        }
+        this.args.change(selectedDate);
+      },
+      "disable": [
+        function (date) {
+          return course.startDate.getUTCDay() !== date.getUTCDay();
```
stopfstedt, line 49: ah, here it is. less kludgy for sure.

dartajax:
I'm trying to put a change request in but don't see that available as an option for some reason so please review this when you get a chance. The action shown below should dismiss the calendar or there should be that "x" available to dismiss the calendar. I understand you can click the calendar on the "7" to confirm but realistically you just want the calendar to go away. You may not be the user who should be confirming the date's correctness.

image

jrjohnson:
@dartajax I wasn't able to add an "X" to close, but I fixed it so when you click in the input a second time the picker will close. let me know what you think.

dartajax:
I like the ability to click back into the field and get the calendar to go away. That is actually a new feature. 

Here is a screen shot of the old date time picker for reference. Notice that there is no drop-down arrow indicator on the month field but if you click it, you can change the month - same thing with year - no obvious indicator but the control made it easy to switch month and year. I like it this way.

image

The new one is shown below with a comment from me in the image.

image

dartajax:
It is possible to delete the year entirely. Of course it will save to whatever the previously saved or selected value was. 

image

Is it possible to have the year be selectable in the same manner as the month and the entire thing centered like in the old version? 

June 2020 - totally centered

image

May 2019 - not centered for example

image

jrjohnson:
I centered the month and year, but I can't change the year input to a dropdown. It would either need to include every year for all of time or have some options to limit the years included and that isn't possible with the library we're using.

jrjohnson:
@dartajax I centered the values (as close as I could) and adjusted the margins so I think the month and year now look like they are the same.

dartajax:
For months where the first of the month is a Sunday, the following occurs ...

image

I'm not sure if I care very much, but I wanted to point it out.

jrjohnson:
I'm not going to dig too far into that, it seems very likely to be something we can't fix.

PR #1398, opened 2020-05-18
Is-date validator

stopfstedt:
i need this in the frontend.

_side-note_: i'm fully acknowledging that we're moving away from moment, but `Date.parse()` is hot garbage and is not usable for production, so `moment.isValid()` is the next go-to we currently have.


stopfstedt:
i'm adding this to frontend directly.

PR #1405, opened 2020-05-27
filter out unpublished/scheduled/blanked prework.

stopfstedt:
partial fix for ilios/frontend#5354


tests/integration/components/dashboard-agenda-test.js
```diff
@@ -47,6 +53,48 @@ module('Integration | Component | dashboard agenda', function (hooks) {
...
+        startDate: today.format(),
+        endDate: tomorrow.format(),
+        location: "room 111",
+        ilmSession: true,
+        slug: 'whatever',
+        postrequisiteSlug: 'something',
+        postrequisiteName: 'first',
+        isPublished: true,
+        isScheduled: true,
+        isBlanked: false,
```
jrjohnson, line 97: I like it when adding tests just means adding fixtures, the tests are already working!

PR #1408, opened 2020-05-29
Add workflow to update dependencies

jrjohnson:
Unfortunately dependabot doesn't update our lock file when transitive
dependencies are updated. This removed the lock file and re-creates it
with fresh data.


.github/workflows/update-transitive-dependenies.yaml
```diff
@@ -0,0 +1,35 @@
+name: Update Transitive Dependencies
+
+on:
+  schedule:
+    - cron: '15 7 * * 1' # weekly, on Monday morning
```
stopfstedt, line 5: nicely timed to go with the morning coffee ;)

PR #1413, opened 2020-06-02
Add polyfill for safari 12

### LINKED ISSUE #1412: Error: Intl.PluralRules is not available in this environment.
We need to add a pollyfill to support safari 12. 
Sentry Issue: [FRONTEND-RB](https://sentry.io/organizations/ilios/issues/1704228892/?referrer=github_integration)

```
Error: Intl.PluralRules is not available in this environment.
Try polyfilling it using "@formatjs/intl-pluralrules"

  at _default (addon-tree-output/moment/lib.js:22:1)
  at _default (addon-tree-output/moment/lib.js:22:1)
  at format (addon-tree-output/moment/lib.js:22:1)
  at compute (addon-tree-output/ember-intl/-private/formatters/format-message.js:68:1)
  at fn (@ember/-internals/glimmer/index.js:2971:1)
...
(27 additional frame(s) were not displayed)
```

jrjohnson:
This version, which is still in active use, doesn't have the Intl libs
we need. ~Using the polyfill.io service creates a script that uses the
browser agent to decide what data to send so users of more modern
browsers won't need to download and execute this.~

Instead of relying on a third party service which may block loading
indefinitely if it is down and which we'll need to remember to keep
updated this loads the polyfills we need for Safari 12 by doing feature
detection and then async loading each script.

Fixes #1412

### DIFF
--- a/addon/utils/load-polyfills.js
+++ b/addon/utils/load-polyfills.js
@@ -1,6 +1,8 @@
 export async function loadPolyfills() {
-  await intlPluralRules();
-  await intlRelativeTimeFormat();
+  await Promise.all([
+    intlPluralRules(),
+    intlRelativeTimeFormat(),
+  ]);
 }
 
 async function intlPluralRules() {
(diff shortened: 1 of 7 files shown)


addon/utils/load-polyfills.js
```diff
@@ -0,0 +1,32 @@
+export async function loadPolyfills() {
+  await Promise.all([
```
stopfstedt, line 2: just curious why we're using native Promise instead of RSVP here. is the plan to phase that libary out?

jrjohnson: No reason. I've been trying to remember to use it when I can since `RSVP` is eventually going to go away. This might be the first time I actually remembered. 😄

PR #1416, opened 2020-06-03
adds controls for adding learners to ILMs

### LINKED ISSUE #1394: Learner Groups Selector Not Appearing After ILM Selector Is Toggled On
Currently, the screen requires a hard refresh before the Learner Group selection tool becomes available. This should not be the case.

Here is a session ...
image

Here is the same session after turning the ILM toggle to the "on" position ...

image

Even if you can't see it in the screen shot, the Learner Group selection area should be available to assign learner groups to this ILM immediately after designating this session to ILM status.

After hard refresh, the Learner Group and Instructor (Group) selectors are now available between Session Administration and Objectives. This is where they should appear directly after toggling the ILM switch - not requiring any additional screen refreshing. 

image

### LINKED ISSUE #1415: Add ability to add individual learners to ILMs

stopfstedt:
this revamps learner group assignment while at it.

fixes #1415

also fixes #1394.


addon/components/collapsed-learnergroups.hbs
```diff
@@ -1,40 +0,0 @@
-<section
```
jrjohnson, line 1: No longer needed? Doesn't work with new learners? Not sure where this went.

addon/components/learnergroup-tree.js
```diff
@@ -2,14 +2,18 @@ import Component from '@glimmer/component';
...
+import { isEmpty, isPresent } from '@ember/utils';
 import {restartableTask} from "ember-concurrency-decorators";
 
 export default class LearnergroupTree extends Component {
   @tracked isHidden = true;
   @tracked selectable = false;
   @tracked hasChildren = false;
 
+  get isRoot() {
+    return isPresent(this.args.isRoot) ? this.args.isRoot : true;
```
jrjohnson, line 14: Is `isPresent` needed here? Seems like this is either `true` or falsey (`false`, `undefined`).

PR #1428, opened 2020-06-08
Use JSON:API when getting data

jrjohnson:
No optimizations or improvements to data loading are added here, just converts the ember-data layer to use the newly created JSON:API.


jrjohnson:
This has been click tested from https://github.com/ilios/frontend/pull/5376 and this side is ready to be merged.

PR #1501, opened 2020-07-07
Trim whitespace from URL input

### LINKED ISSUE #1500: Left Trim On URL To Help With Validation - Offerings Editor
During my initial testing on this just now, I accidentally pasted in something like the following " https://meetinglink.url" - the important part being the blank space NBSP; at the beginning of the URL. Other than that, the URL was fine. Can we perform a left trim if there are blank spaces at the beginning of a bad copy and paste job?

If not, no problem. It might be helpful.

jrjohnson:
This makes it easier to paste into the form and pass validation.

Fixes #1500

### DIFF
--- a/addon/components/offering-form.js
+++ b/addon/components/offering-form.js
@@ -246,6 +246,7 @@ export default class OfferingForm extends Component {
 
   @action
   changeURL(value) {
+    value = value.trim();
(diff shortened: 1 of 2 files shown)


dartajax:
Nice work.

PR #1504, opened 2020-07-09
Improve screen position when visiting session

jrjohnson:
When clicking on a session further down the list of sessions in a course
users were being left somewhere in the middle of the session instead of
at the top. This seems to be caused by the scroll working while the page
is still building. Waiting until the next tick of the runloop results in
the desired position of the screen.

I also moved this up a level because I think it gives a better user
experience to retain the return to session link.


dartajax:
Looks good to me as well.

PR #1540, opened 2020-07-23
rewires session description for API v3

stopfstedt:
companion PR to https://github.com/ilios/ilios/pull/2989

refs https://github.com/ilios/ilios/pull/2917


addon/components/session-overview.js
```diff
@@ -34,7 +34,6 @@ export default class SessionOverview extends Component {
   @tracked showSpecialAttireRequired = false;
   @tracked showSpecialEquipmentRequired = false;
   @tracked isIndependentLearning = false;
-  @tracked description;
```
jrjohnson, line 37: Since this is still modified shouldn't it still be tracked?

stopfstedt: correct, i should have made this a separate commit. turns out this property was declared twice, i'm removing the duplicate here. see:
 https://github.com/ilios/common/pull/1540/files/67fae8b56b764fd96ceac49e0dedc067201fde0d#diff-1b6a1f0cf1f929becb12d04a77bc8ec6R24 for the remaining declaration.

jrjohnson: Ah, I see it now. Had to look like 5 times!

PR #1561, opened 2020-08-04
prevents double-listing of ILM-based events if they are already shown as post-reqs

### LINKED ISSUE #1553: Activities View and WaaG Need To Follow Correct Logic for Displaying ILM's with Pre-Work
The following is a paste from a student calendar ...

image

"Formative 1" is an ILM that is due before another ILM called "Formative 1 Due By 8/5 at 8:00 AM".

The upper part of the screen with the Pre-Work is correct. See below.

image

stopfstedt:
this filters out any ILMs-based events from the agenda if they are already listed as post-requisite in the "pre-work" section.

Selection_073

fixes #1553


stopfstedt:
PR withdrawn as this does not address the issue at hand. see https://github.com/ilios/ilios/pull/3040 for the actual solution.

PR #1570, opened 2020-08-07
respect sort order input if its value is false.

### LINKED ISSUE #1569: Sortable Table Header component - icon does not get updated on sort order reveral.
At first, I was like this.

Selection_096

But then, ... still like this.

Selection_095

fix this.

stopfstedt:
fixes #1569

### DIFF
--- a/addon/components/sortable-th.js
+++ b/addon/components/sortable-th.js
@@ -11,7 +11,7 @@ export default class SortableTh extends Component {
   }
 
   get sortedAscending() {
-    return this.args.sortedAscending || true;
+    return this.args.sortedAscending ?? true;
   }
 
   get sortedBy() {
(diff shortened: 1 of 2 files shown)


addon/components/sortable-th.js
```diff
@@ -11,6 +11,9 @@ export default class SortableTh extends Component {
   }
 
   get sortedAscending() {
+    if (false === this.args.sortedAscending) {
+      return this.args.sortedAscending;
+    }
     return this.args.sortedAscending || true;
```
jrjohnson, line 17: I think this was supposed to be `return this.args.sortedAscending ?? true;` which would return the value if set or else return true and should work without the addition to check for explicit false.

stopfstedt: good point. replaced this with the suggested shorthand form.

PR #1588, opened 2020-08-18
Replace our URL validation library

jrjohnson:
This is a more maintained validation system (and we're already using it
on the frontend). Unfortunately it still doesn't catch a bad slash in the
query param so I've added a test that is allowed to fail and will
attempt a PR upstream to fix.

Refs ilios/ilios#3050

### DIFF
--- a/tests/integration/components/offering-form-test.js
+++ b/tests/integration/components/offering-form-test.js
@@ -1,4 +1,4 @@
-import { module, test } from 'qunit';
+import { module, test, todo } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import { setupMirage } from 'ember-cli-mirage/test-support';
 import { render } from '@ember/test-helpers';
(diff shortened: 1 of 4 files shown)


tests/integration/components/offering-form-test.js
```diff
@@ -442,4 +442,11 @@ module('Integration | Component | offering form', function(hooks) {
     await component.url.set('  http://example.com  ');
     assert.equal('http://example.com', component.url.value);
   });
+
+  todo('rejects query param with trailing slash ilios/ilios#3050', async function(assert) {
+    await render(hbs`<OfferingForm @close={{noop}} @showRoom={{true}} />`);
+    assert.notOk(component.url.hasError);
+    await component.url.set('http://example.com?jayden=awesome/');
```
stopfstedt, line 449: :dog: :heart:

PR #1593, opened 2020-08-21
show campus name of record where display name diverges.

stopfstedt:
Selection_126
Selection_127
Selection_139
Selection_138
Selection_141
Selection_142
Selection_143

out of scope:
- course print view - we don't want/need tooltips in here, this is mean for print. 
- learning material sort manager: this is minor information there, adding tooltips in this sorting interface would add just add noise.
- learning material search: tool-tiping search results, no bueno. 
- offerings details in session grid: we run an abbreviated, comma separated list of user names there. again, adding tooltips here is too much, let's keep it to where user assignment is happening
- course instructor visualization: adding a tooltip within a tooltip, not desirable.
- the various user searches: in-scope for this, but will be handled as separate issue

refs https://github.com/ilios/frontend/issues/5524

### DIFF
--- a/addon/components/detail-instructors-list.hbs
+++ b/addon/components/detail-instructors-list.hbs
@@ -1,5 +1,6 @@
 <div
   class="detail-instructors-list"
+  data-test-instructors-list
 >
   {{#if @instructors.length}}
     <h4>
@@ -9,7 +10,7 @@
     <ul class="instructors-list" data-test-instructors>
       {{#each (sort-by "fullName" @instructors) as |user|}}
         <li>
-          {{user.fullName}}
+          <UserNameInfo @user={{user}} />
         </li>
       {{/each}}
     </ul>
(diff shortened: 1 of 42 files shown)


addon/components/detail-instructors-list.hbs
```diff
@@ -9,7 +10,29 @@
     <ul class="instructors-list" data-test-instructors>
       {{#each (sort-by "fullName" @instructors) as |user|}}
         <li>
-          {{user.fullName}}
+          <span data-test-name>
```
jrjohnson, line 13: This chunk of template seems to be repeated a lot. The only difference I can spot is that `data-test` attribute on the full name span. Wouldn't making this a component simplify the mouse over registration for the tooltip as well as removing duplicate template and JS code?

stopfstedt: all switched out and consolidated. i did not experience any of the performance degradation i was worried about.

jrjohnson (approved):

PR #1617, opened 2020-09-08
Improve style and accessibility of week at a glance

### LINKED ISSUE #1614: Font Size Reduction On Week At A Glance / All Weeks
This is a mockup from ilios/ilios#3104 and this issue is related loosely to that one. 

Week At A Glance 

image

All Weeks

image

jrjohnson:
Get the font sizes under control, mostly this was a problem of applying
the .title class way to many times. Restructuring the HTML I also added
some a11y tests.

Fixes #1614

### DIFF
--- a/addon-test-support/ilios-common/page-objects/components/week-glance.js
+++ b/addon-test-support/ilios-common/page-objects/components/week-glance.js
@@ -4,12 +4,14 @@ import {
   text
 } from 'ember-cli-page-object';
 import weekGlanceEvent from './week-glance-event';
+import weekGlancePreWork from './week-glance-pre-work';
 
 const definition = {
   scope: '[data-test-week-glance]',
   title: text('[data-test-week-title]'),
   offeringEvents: collection('[data-test-week-glance-event]', weekGlanceEvent),
   ilmEvents: collection('[data-test-ilm-events] [data-test-week-glance-event]', weekGlanceEvent),
+  preWork: collection('[data-test-ilm-pre-work] [data-test-week-glance-pre-work]', weekGlancePreWork),
 };
 
 export default definition;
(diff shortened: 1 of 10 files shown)


stopfstedt (approved):
LGTM.

stopfstedt:
@dartajax please give this a visual review, thanks!

dartajax:
very nice but please review my comment below 

image

jrjohnson:
Good call @dartajax, more space added. How's that?

dartajax:
I will approve this and merge either way in all likelihood but I am not seeing any space between "Pre-work" and the first session "Neoplasia ..." and the larger fonts are back. I liked the indentation too.

image

dartajax:
image

dartajax:
Suggestion ... re-use this component as it appears underneath the calendar - that way it can be maintained in the same place and appear the same to users in both locations.

image

dartajax:
That component I pasted in above reflects generally the way I believe the fonts should be sized and styled. That's just my opinion though. Fonts shouldn't be jumping around in size.

jrjohnson:
@dartajax I'm not seeing the font size issues you're seeing. What browser, demo user, and URL is that? 
I adjusted the pre work to not take up empty space when it isn't present so that should be fixed.

dartajax:
The fonts look better and pre-work is nice. Great work. One final suggestion - can we move everything in a little bit, Right now everything is flush against the left wall - as far left as possible with zero padding.

image

dartajax:
everybody move over two spaces to the right

jrjohnson:
That space will show up when this is placed in the frontend as the outside margin is provided there.

Looks like:
Screen Shot 2020-09-10 at 9 20 06 AM

PR #1629, opened 2020-09-11
check against report owning school, not linked program's owning school.

stopfstedt:
this fixes a discrepancy between the permissions checker implementations in backend and common.

realigned with what the backend validator by checking the school property of the correct entity: 

https://github.com/ilios/ilios/blob/master/src/RelationshipVoter/CurriculumInventoryExport.php#L43

refs https://github.com/ilios/frontend/issues/5570

### DIFF
--- a/addon/services/permission-checker.js
+++ b/addon/services/permission-checker.js
@@ -321,8 +321,7 @@ export default Service.extend({
       return false;
     }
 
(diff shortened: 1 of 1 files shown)


stopfstedt:
this is wrong. the bug must be elsewhere.

PR #1669, opened 2020-09-29
Properly styles and titles navigational elements

stopfstedt:
Selection_231

fixes https://github.com/ilios/common/issues/1296

### DIFF
--- a/addon/components/weekly-events.hbs
+++ b/addon/components/weekly-events.hbs
@@ -1,9 +1,21 @@
 <div class="weekly-events">
   <div class="year">
     <h2>
-      <FaIcon @icon="backward" {{on "click" this.decrementYear}}/>
+      <FaIcon
+        @icon="backward"
+        @title={{t "general.goToPreviousYear"}}
+        role="button"
+        class="clickable link"
+        {{on "click" this.decrementYear}}
+      />
       {{@year}}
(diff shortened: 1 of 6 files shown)


addon/components/weekly-events.hbs
```diff
@@ -1,9 +1,23 @@
...
+<div class="weekly-events" data-test-weekly-events>
+  <div class="year" data-test-top-nav>
     <h2>
-      <FaIcon @icon="backward" {{on "click" this.decrementYear}}/>
-      {{@year}}
-      <FaIcon @icon="forward" {{on "click" this.incrementYear}}/>
+      <FaIcon
+        @icon="backward"
+        @title={{t "general.goToPreviousYear"}}
+        role="button"
```
jrjohnson, line 7: Passing `@click` to the `<FaIcon` will add the necessary a11y markup instead of needing to handle it here.

stopfstedt: sorry to disappoint, but that's incorrect. if no `role` attribute is passed then `role` will default to `img`.
Selection_232

jrjohnson: Oh no! Well that's not good. Ok.

jrjohnson (approved):

PR #1680, opened 2020-10-02
Use a getter to built headers

jrjohnson:
This ensures it will react to changes that happen after this adapter is
constructed.


stopfstedt (approved):
noice.

PR #1683, opened 2020-10-02
puts some guard rails up before interacting with menu element.

stopfstedt:
can't reproduce it locally, but throwing up some extra check won't hurt.

hopefully fixes https://github.com/ilios/common/issues/1236


jrjohnson:
This may have been resolved in https://github.com/ilios/common/pull/1552, but I think we should add this as well.

jrjohnson (approved):

PR #1716, opened 2020-10-19
re-layout date-time/recurrence info in offering form.

stopfstedt:
refs https://github.com/ilios/frontend/issues/3402#issuecomment-682202928

### DIFF
--- a/addon-test-support/ilios-common/page-objects/components/offering-form.js
+++ b/addon-test-support/ilios-common/page-objects/components/offering-form.js
@@ -1,8 +1,18 @@
-import {clickable, collection, create, fillable, isVisible, property, text, value} from 'ember-cli-page-object';
+import {
+  clickable,
+  collection,
+  create,
+  fillable, hasClass,
+  isPresent,
+  isVisible,
+  property,
+  text,
+  value,
+} from 'ember-cli-page-object';
 import { flatpickrDatePicker } from 'ilios-common';
-import learnerSelectionManager from './learner-selection-manager';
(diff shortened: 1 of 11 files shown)


dartajax:
@jrjohnson @saschaben I noticed in this PR that when you create a new offering (well start the process anyway) you are routed to the bottom of page below where you would be adding offering details. We had a ticket in common - https://github.com/ilios/common/issues/899 which made this the case. I entered it. This was probably in a large monitor / one cohort in the course situation. On my laptop with course URL as shown here https://deploy-preview-1716--ilios-common.netlify.app/courses/1749/sessions/58895 go to the bottom and click the "Add Offering" + button and this is the result ...

image

this would probably be better ...

image

stopfstedt:
@dartajax i removed the scroll-to-bottom, relabeled the search boxes and standardized between labels and their corresponding form elements (as much as possible). please have another look, thanks.

dartajax:
Is there any way we can have this ...?

image

Instead of this ...?

image

The user is at the bottom of the form and when there are no offerings yet, is there a way to have the offering form be in the usable (not requiring the user to scroll) part of the screen.

stopfstedt:
@dartajax i brought the scroll-into-view feature back for new offerings, and anchored the scroll target at the top of the form. please have another look, thanks!

dartajax (approved):
This is good.

jrjohnson (approved):

dartajax:
Is the "New Offering" label supposed to appear in the middle of the screen? If there is no way to make it at the top of the screen instead, this seems like the best option so far.

PR #1733, opened 2020-10-29
moves scroll-into-view into wrapper component.

stopfstedt:
@dartajax i moved the scroll-into-view functionality out of the offering form component itself and into the wrapper component for new offerings. the new scroll target is the "New Offering" label.


dartajax:
It looks like I added my comment to the wrong ticket - sorry about that.

dartajax (approved):

PR #1741, opened 2020-11-03
improves URL validation.

### LINKED ISSUE #1740: Need Error or Better Validation of URL >> Course LM
The scenario below includes an apparently active Done button and a Cancel button (which should be active no matter what). the problem here is that clicking the Done button does not do anything and no error message is returned. This would allow the user to gleam the information regarding what is missing. Or have the Done button be disabled until a file is attached - developer's choice. Important concept - will be discussed.

image

stopfstedt:
fixes #1740

### DIFF
--- a/addon/components/new-learningmaterial.hbs
+++ b/addon/components/new-learningmaterial.hbs
@@ -80,12 +80,17 @@
         {{t "general.url"}}:
       </label>
       <span>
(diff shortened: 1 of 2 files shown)


dartajax (approved):
This is a great improvement. LGTM.

PR #1760, opened 2020-11-12
Sort learnergroups in picker

stopfstedt:
Selection_022

fixes https://github.com/ilios/frontend/issues/1776

### DIFF
--- a/addon/components/learnergroup-tree.js
+++ b/addon/components/learnergroup-tree.js
@@ -58,6 +58,7 @@ export default class LearnergroupTree extends Component {
     }
   }
 
+  @action
   sortByTitle(learnerGroupA, learnerGroupB) {
     const locale = this.intl.get('locale');
     return learnerGroupA.title.localeCompare(learnerGroupB.title, locale, {numeric: true});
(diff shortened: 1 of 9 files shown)


addon/components/learnergroup-tree.js
```diff
@@ -54,4 +54,8 @@ export default class LearnergroupTree extends Component {
       this.args.add(learnerGroup);
     }
   }
+
+  sortByTitle(learnerGroupA, learnerGroupB) {
+    return learnerGroupA.title.localeCompare(learnerGroupB.title, undefined, {numeric: true});
```
jrjohnson, line 59: Instead of undefined can you pass the current locale in [like this](https://github.com/ilios/common/blob/7ca9d7ea26810af47af90a19702c709f8328a9b1/addon/components/user-search.js#L126-L131) it may do the same thing by reading the locale from the html language attribute, but I like the explicit one better.

jrjohnson (changes requested):

dartajax:
It looks good from the screenshots but I can't get this working to verify in common - https://deploy-preview-1760--ilios-common.netlify.app/learnergroups

dartajax (approved):
Looks good to me - I approve.

jrjohnson (approved):

PR #1773, opened 2020-11-17
Give course title input field more space

### LINKED ISSUE #1772: Improve Course Title Editing Box
image

related to https://github.com/ilios/common/issues/1770

image

stopfstedt:
fixes #1772 

Selection_046

### DIFF
--- a/addon/components/course-header.hbs
+++ b/addon/components/course-header.hbs
@@ -26,16 +26,16 @@
         {{/each}}
       </EditableField>
     {{else}}
-      <h2>
+      <h4>
         {{#if @course.locked}}
           <FaIcon @icon="lock" />
         {{/if}}
         {{@course.title}}
-      </h2>
+      </h4>
     {{/if}}
-    <h3>
+    <h4 class="academic-year">
       {{@course.academicYear}}
-    </h3>
+    </h4>
   </span>
   <span class="course-publication">
     {{#if @editable}}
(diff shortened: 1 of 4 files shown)


addon/components/course-header.hbs
```diff
@@ -26,16 +26,16 @@
         {{/each}}
       </EditableField>
     {{else}}
-      <h2>
+      <h4>
```
jrjohnson, line 29: I know we don't enforce it everywhere, but the header level here *should* be 2 as it is the critical data on the page.

jrjohnson (changes requested):
Not sure about this solution, why would the title area only be bigger when editing, shouldn't it take up the same space as the full title did because it contains the same characters?

dartajax (approved):
It works for me.

dartajax:
Sao @jrjohnson's issue (and I can understand it) is that the font sizes are different between display mode and edit mode. Is that it?

addon/components/course-header.hbs
stopfstedt: handled.

stopfstedt:
> Not sure about this solution, why would the title area only be bigger when editing, shouldn't it take up the same space as the full title did because it contains the same characters?

@jrjohnson would this be better then?

Selection_052

jrjohnson (approved):

jrjohnson:
Egad, I really don't like that. I was worried about the year jumping around when editing that it might be distracting if it changed lines, but I didn't trust the CSS.  I just played with it a bunch and it works really well. I take back my objection, this is perfect.

PR #1778, opened 2020-11-18
re-styles selected learner groups as removable tag list mixin.

### LINKED ISSUE #1776: Remove Padding From Learner Group Labels in Offerings Editor
Here is how it currently is ...

image

Something like ...

image

This is so more information can fit in a smaller space requiring less scrolling required for the user to obtain access to the action buttons Done / Cancel.

Similar badge / label sizes being suggested here can be seen also in the newly revamped selected instructor / instructor group section ...

image

stopfstedt:
this is a space-saving measure.

fixes #1776 

Selection_054


dartajax (approved):
very nice 🥇

Issue #1822, opened 2020-12-10
User Should Be Sent to Bottom of Screen During Session Copy Action


dartajax:
I want to make a copy of the following session as shown below.

![image](https://user-images.githubusercontent.com/7493331/101837554-0ed1e500-3af4-11eb-8454-82491920b7cc.png)

After clicking as shown above, the screen changes into the following ...

![image](https://user-images.githubusercontent.com/7493331/101839277-f0211d80-3af6-11eb-9170-dd2acdceb576.png)

The user wold have to know they would need to scroll down to the session copy details area. They should be taken to the bottom of the form when they click the `copy session` button.

![image](https://user-images.githubusercontent.com/7493331/101839392-278fca00-3af7-11eb-91e3-02a2f461fdd9.png)

PR #1828, opened 2020-12-11
scrolls session copy form into view.

stopfstedt:
fixes https://github.com/ilios/common/issues/1822

### DIFF
--- a/addon/components/session-copy.hbs
+++ b/addon/components/session-copy.hbs
@@ -1,6 +1,7 @@
 <div
   class="session-copy"
   {{did-insert (perform this.setup) @session}}
+  {{did-insert this.scrollIntoView}}
   {{did-update (perform this.setup) @session}}
 >
   <div class="backtolink">
--- a/addon/components/session-copy.hbs
+++ b/addon/components/session-copy.hbs
@@ -1,7 +1,6 @@
 <div
   class="session-copy"
   {{did-insert (perform this.setup) @session}}
-  {{did-insert this.scrollIntoView}}
   {{did-update (perform this.setup) @session}}
 >
   <div class="backtolink">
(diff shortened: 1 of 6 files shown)


addon/components/session-overview.js
```diff
@@ -316,4 +317,10 @@ export default class SessionOverview extends Component {
 
     yield this.args.session.save();
   }
+
+  @action
+  transitionToCopy() {
+    this.router.transitionTo('session.copy', this.args.session);
```
jrjohnson, line 323: I'm not necessarily opposed to this solution, but it would seem easier to control this from inside the session-copy component itself with `did-insert scroll-into-view`.

jrjohnson (changes requested):

addon/components/session-overview.js
stopfstedt: good point. fixed. please re-review, thanks.

addon/components/session-copy.hbs
```diff
@@ -1,21 +1,22 @@
 <div
   class="session-copy"
   {{did-insert (perform this.setup) @session}}
+  {{did-insert this.scrollIntoView}}
```
jrjohnson, line 4: Ugh, bad initial suggestion, sorry. We actually have a modifier that does this. You can just do `{{scroll-into-view}}` and it should handle it.

jrjohnson (changes requested):

addon/components/session-copy.hbs
stopfstedt: > Ugh, bad initial suggestion, sorry. We actually have a modifier that does this. You can just do `{{scroll-into-view}}` and it should handle it.

haha, no prob. done. and good to know!

jrjohnson (approved):

PR #1829, opened 2020-12-12
Submit login form on Enter

stopfstedt:
copy your token, press Enter, done!

the rest is just cleanup.

### DIFF
--- a/tests/dummy/app/templates/login.hbs
+++ b/tests/dummy/app/templates/login.hbs
@@ -2,7 +2,7 @@
 <input
   type="text"
   value={{this.jwt}}
-  {{on "input" this.updateJwt}}
+  {{on "input" (pick "target.value" (set this.jwt))}}
   {{on "keypress" (perform this.loginOnEnter)}}
   disabled={{this.loginOnEnter.isRunning}}
 />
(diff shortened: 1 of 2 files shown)


tests/dummy/app/templates/login.hbs
```diff
@@ -2,8 +2,15 @@
 <input
   type="text"
   value={{this.jwt}}
-  {{on "input" this.updateJwt}} />
-
-<button type="button" {{on "click" this.login}}>{{t "general.login"}}</button>
+  {{on "input" (pick "target.value" (set this.jwt))}}
+  {{on "keypress" (perform this.loginOnEnter)}}
+  disabled={{this.loginOnEnter.isRunning}}
```
jrjohnson, line 7: I feel like this should be disabled whenever `login` is running, not just on enter.

jrjohnson (changes requested):

tests/dummy/app/templates/login.hbs
stopfstedt: you're right. i corrected this.

jrjohnson (approved):

dartajax:
Works for me. 💯

Issue #1838, opened 2020-12-16
Date Pickers Aren't Showing Up - Maybe Present But Behind The Scenes


dartajax:
To reproduce, 

* Find a date picker - Course End Date will be used in this example 
* Click into the control 
* Notice that the date picker should be available to select the date 
* Click again and again - still no picker is available
* Go to another tab on your browser and review something else 
* Return to the tab on which Ilios is fired up and ..
* Enjoy the glory of the date picker which is now available 
* Clicking again will cause the date picker to disappear
* Click again and it will re-appear as it should 

The date picker is now operating normally but only after having to go to another tab and return to Ilios. This is sure to cause user frustration and chagrin.

PR #1839, opened 2020-12-16
removes click-event handler

stopfstedt:
it was making the date picker too trigger happy.
turns out it is not necessary in order to close the picker.

tested with latest FF and Chrome on Linux.

fixes https://github.com/ilios/common/issues/1838


dartajax:
I tested this in Chrome, Safari, and Opera on my older Mac and it seems to work fine. LGTM. 💯

jrjohnson (approved):

PR #1882, opened 2021-01-14
increases number of available course level options.

### LINKED ISSUE #1881: Course Level Should Be Increased From MAX of 5 up to 10
Currently, Ilios is hard wired to make it so a Course Level must be selected between 1 and 5. This needs to be increased to 10.

image

stopfstedt:
fixes #1881


jrjohnson (approved):
I did my best to understand this change, but given the complexity and number of lines and files touched I'm not sure my review was very thorough 😜

dartajax (approved):
That's not the trickiest chunk of code I've ever seen. It works fine in any case.

PR #1884, opened 2021-01-14
Prevent editing course when rolling over

### LINKED ISSUE #1883: Disable Course Level Fields During Rollover
Users are occasionally confused when rolling over a course since the fields in the Course detail at the top of the screen are still enabled. Those fields are not part of the rollover process and out of context while performing a Course rollover. Worse yet, they may think the fields have something to do with the Course being created in the rollover process rather than modifying the original source Course, which is what happens if those fields get edited in this context. It's better to lock the fields to prevent anything from going awry.

image

jrjohnson:
This is confusing as users may believe they're modifying the rollover
process and not the source course. Fixes #1883

### DIFF
--- a/addon/components/ilios-course-details.hbs
+++ b/addon/components/ilios-course-details.hbs
@@ -10,8 +10,8 @@
       </LinkTo>
     </div>
   {{/if}}
-  <CourseHeader @course={{@course}} @editable={{@editable}} />
-  <CourseOverview @course={{@course}} @editable={{@editable}} />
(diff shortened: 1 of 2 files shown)


stopfstedt (approved):
slick!

dartajax (approved):
fine with me - good stuff

dartajax:
weird - I couldn't merge this because I thought it was already in "ready for review" mode - @stopfstedt and I have approved this - anyway I'm gonna merge it 'cuz I like it

PR #1898, opened 2021-01-25
show pointer cursor on calendar event only if event is clickable.

### LINKED ISSUE #1894: non-clickable calendar events should have different pointer
here's a screenshot of an calendar event in the user admin calendar.

Screenshot from 2021-01-22 16-14-46

it is a not clickable, however the mouse pointer indicates otherwise. this is misleading.

style the cursor for non-clickable events differently.

stopfstedt:
fixes #1894 

needs a visual review.

### DIFF
--- a/app/styles/ilios-common/components/daily-calendar-event.scss
+++ b/app/styles/ilios-common/components/daily-calendar-event.scss
@@ -1,5 +1,6 @@
 .daily-calendar-event {
   @include ilios-button-reset;
+  cursor: default;
   border-radius: $base-border-radius;
   display: flex;
   flex-direction: column;
@@ -13,6 +14,9 @@
   overflow: hidden;
   z-index: 2;
 
+  &.clickable {
+    cursor: pointer;
+  }
 
   .ilios-calendar-event-time {
     color: $maroon;
--- a/app/styles/ilios-common/components/monthly-calendar.scss
+++ b/app/styles/ilios-common/components/monthly-calendar.scss
@@ -65,12 +65,17 @@
       }
 
       .month-event {
+        cursor: default;
         border: 1px solid $grey;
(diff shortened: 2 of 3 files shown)


dartajax (approved):
I couldn't really test the admin console user calendar with a common netlify build. Gonna approve anyway. Code looks good though.

app/styles/ilios-common/components/daily-calendar-event.scss
```diff
@@ -13,6 +14,9 @@
   overflow: hidden;
   z-index: 2;
 
+  &.clickable {
```
jrjohnson, line 17: This shouldn't be necessary since it's defined globally in https://github.com/ilios/common/blob/f6e464e47c35899a655c4cd5b799efa1656a14a5/app/styles/ilios-common/components/body.scss#L82

jrjohnson (changes requested):

app/styles/ilios-common/components/daily-calendar-event.scss
stopfstedt: image

unfortunately, the explicit CSS declaration here is necessary, the `.daily-calendar-event` overwrites `.clickable`. they have the same specificity, but the event class appears after the generic clickable style. 
thinking of alternatives, we could throw a `.non-clickable` style in the class attribute, or perhaps use a different element  that's not a `<button>` for putting those events on the calendar.

stopfstedt: image

unfortunately, this extra style declaration is necessary. `.clickable` gets overwritten otherwise.

alternatives - none of which i like better: 
- add an explicit `non-clickable` class to the button
- use a `div` instead of a `button`.

jrjohnson (approved):

PR #1905, opened 2021-01-28
show end-date in calendar event details.

### LINKED ISSUE #1904: End Time Needs to Be Included in Event Detail
See below - we have the start time but not the end time. This results in more user clicks. They have to go back to the calendar to figure out the duration of the event they are attending. We don't need duration here but we do need the end time - user group suggestion.

image

stopfstedt:
fixes #1904 

start/end date are the same:

startend_same_date

start/end date fall on the same calendar day:
startend_same_day

start/end date fall on different calendar days:
startend_different_days


dartajax (approved):
I like it. Great work.

jrjohnson (approved):

PR #1908, opened 2021-01-29
Update Ember CLI to  3.24

jrjohnson:
This includes a blueprint update to use prettier to lint and format JS files so there I significant churn from the first auto fix run.


stopfstedt (approved):
LGTM

PR #1925, opened 2021-02-05
removes obsolete "academicYearTitle" property from academic-year model

stopfstedt:
refs ilios/ilios#3272

### DIFF
--- a/addon/models/academic-year.js
+++ b/addon/models/academic-year.js
@@ -1,5 +1,5 @@
 import Model, { attr } from '@ember-data/model';
 
 export default Model.extend({
-  title: attr('number'),
+  title: attr('string'),
 });
--- a/addon/models/academic-year.js
+++ b/addon/models/academic-year.js
@@ -1,9 +1,5 @@
 import Model, { attr } from '@ember-data/model';
-import { computed } from '@ember/object';
 
 export default Model.extend({
   title: attr('number'),
-  academicYearTitle: computed('title', function () {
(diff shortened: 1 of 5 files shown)


jrjohnson (changes requested):
Should come with a minor API version bump to `3.3` to avoid UI confusion relying on this property without the correct backend version.

stopfstedt:
> Should come with a minor API version bump to `3.3` to avoid UI confusion relying on this property without the correct backend version.

good points. handled. :heavy_check_mark:

jrjohnson (dismissed):

addon/models/academic-year.js
```diff
@@ -1,9 +1,5 @@
 import Model, { attr } from '@ember-data/model';
-import { computed } from '@ember/object';
 
 export default Model.extend({
   title: attr('number'),
```
jrjohnson, line 4: This is a string now.

jrjohnson (changes requested):

addon/models/academic-year.js
stopfstedt: :heavy_check_mark:

dartajax: Is this all sorted out? I'd be glad to merge if so.

stopfstedt: @dartajax  this needs the corresponding backend change to land first: https://github.com/ilios/ilios/pull/3306

jrjohnson (approved):

PR #1935, opened 2021-02-11
Lock ember version

jrjohnson:
One of the new versions has an issue with strings that we don't want to
handle. Will track down and report, but for now this is a fix.


stopfstedt (approved):
SUPERAPPROVED

PR #1964, opened 2021-02-22
show year range or single year in academic year dropdown in course rollover dialog

stopfstedt:
refs https://github.com/ilios/ilios/issues/3272
refs https://github.com/ilios/frontend/issues/5866

image

image

### DIFF
--- a/addon/components/course-rollover.js
+++ b/addon/components/course-rollover.js
@@ -12,15 +12,13 @@ export default class CourseRolloverComponent extends Component {
   @service store;
   @service flashMessages;
   @service iliosConfig;
+
   @Length(3, 200) @NotBlank() @tracked title;
   @NotBlank() @tracked selectedYear;
-
   @tracked years;
-  @tracked selectedYear;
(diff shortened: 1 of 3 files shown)


addon/components/course-rollover.js
```diff
@@ -46,6 +45,9 @@ export default class CourseRolloverComponent extends Component {
       },
     });
     this.changeSelectedYear(this.years.firstObject);
+    this.academicYearCrossesCalendarYearBoundaries = yield this.iliosConfig.itemFromConfig(
```
jrjohnson, line 48: I think this needs to move up above `allCourses` otherwise there could be a flash of wrong values while the config is loading and the yield is working.

jrjohnson (changes requested):

addon/components/course-rollover.js
stopfstedt: good catch :heavy_check_mark:

jrjohnson (approved):

PR #1965, opened 2021-02-22
Add a default value to resolve async helper

jrjohnson:
This allows passing an array or null in to ensure some expected value
will exist when this is `@use`ed in another component and alleviating the
need for an interim getter.

Wip:
- [x] Convert existing usage of this helper to use the default value where needed


stopfstedt (approved):
LGTM

PR #1994, opened 2021-03-04
Refactor user search

jrjohnson:
Pulling out the proxy objects and restructuring into several components
makes this usable with resources and other async data strategies.


stopfstedt (approved):
super nice!

PR #1995, opened 2021-03-05
Build and test with embroiderer

jrjohnson:
This is the upcoming build system for ember apps and we'd like to start
using it in the frontend, we should test here to ensure we are
compatible.


stopfstedt (approved):
LGTM

PR #1997, opened 2021-03-07
Only save materials where sort order has changed

jrjohnson:
This speeds up sorting significantly in a large list and hopefully makes
the wait-saving indicator unnecessary.

Refs https://github.com/ilios/frontend/issues/5900


stopfstedt (approved):
LGTM

PR #2022, opened 2021-03-19
Add autofocus modifier

jrjohnson:
This allows us to place focus easily within an element as soon as it is
inserted.


stopfstedt (approved):
awesome.

Issue #2041, opened 2021-04-02
[META] rewire event handling for components


stopfstedt:
passing event handlers directly now throws a template linter error. it's relatively easy to work around this.

- [x] `SortableHeader` component - replace `@click` with `@onClick`
- [x] `SortableTh` component - replace `@click` with `@onClick`
- [x] `DatePicker` component - replace `@change` with `@onChange`

for reference, see https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-passed-in-event-handlers.md

stopfstedt:
there are probably others that need this treatment, update the list above as those get identified.

PR #2044, opened 2021-04-04
[breaking] change attr name for passing in a click or change event handler.

stopfstedt:
refs #2041 

_this constitutes a breaking change and warrants a major version bump_


jrjohnson (approved):
👍

PR #2057, opened 2021-04-13
give grid columns more room to breathe.

### LINKED ISSUE #2055: course overview: editinplace controls don't fit into their space
space these controls out a bit more. 

image

stopfstedt:
fixes #2055 

since this is a global style change, i  reviewed and included screenshots of affected components/areas in both common and frontend

course-overview
course-summary-header
program-overview
report-overview
sequence-block-overview
session-overview

### DIFF
--- a/app/styles/ilios-common/mixins/ilios-overview.scss
+++ b/app/styles/ilios-common/mixins/ilios-overview.scss
@@ -44,6 +44,6 @@
 
   @include for-tablet-and-up {
     grid-row-gap: .5rem;
-    grid-template-columns: repeat(auto-fill, minmax(300px, 1fr));
+    grid-template-columns: repeat(auto-fill, minmax(400px, 1fr));
   }
 }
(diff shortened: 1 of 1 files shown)


dartajax (commented):
image


image

stopfstedt:
@dartajax  see https://github.com/ilios/common/pull/2058

jrjohnson (changes requested):
This will bump small tablets down to a one column layout. That isn't a big problem, but I think we probably need to split this into a query for tablets that evenly decides into two 350px columns and change this to use [for-laptop-and-up](https://github.com/ilios/common/blob/master/app/styles/ilios-common/mixins/media.scss#L13) which has more room 3+ 400px columns.

stopfstedt:
> This will bump small tablets down to a one column layout. That isn't a big problem, but I think we probably need to split this into a query for tablets that evenly decides into two 350px columns and change this to use [for-laptop-and-up](https://github.com/ilios/common/blob/master/app/styles/ilios-common/mixins/media.scss#L13) which has more room 3+ 400px columns.

handled. 

for reference, here is how the course overview looks now on tablets:

image
 
i checked all other areas that use those styles again, and the layout works there as well.

jrjohnson (approved):

dartajax (approved):

PR #2060, opened 2021-04-13
corrects name of session objectives attribute.

### LINKED ISSUE #2059: publish all sessions: objectives count is incorrect
_the one that got away_

this is leftover from the objectives-related changes last year.

currently, it only shows "no objectives", regardless of whether session objectives exist or not:

wrong

should be:

right

fix this.

stopfstedt:
fixes #2059


jrjohnson (approved):
Nice! It's like 200% more test covered now.

PR #2128, opened 2021-05-14
adds controls for setting hours and minutes on ILM duedate.

stopfstedt:
fixes https://github.com/ilios/common/issues/2127


dartajax:
I can see us discussing the formatting; but for an initial pass at it, I like it.

dartajax (approved):
It looks good to me - a little strange in phone mode but still useable without issue.

addon/serializers/ilm-session.js
```diff
@@ -1,15 +1,8 @@
...
   serialize(snapshot, options) {
     const json = super.serialize(snapshot, options);
-    // set time to 5pm, always.
-    const dueDate = moment(json.data.attributes.dueDate);
-    dueDate.hour('17');
-    dueDate.minute('00');
-    json.data.attributes.dueDate = dueDate.format();
-
     //don't persist this, it is handled by the server
     delete json.data.attributes.updatedAt;
```
jrjohnson, line 7: `ilmSession` doesn't have an `updatedAt` property, I'm guessing this is an ancient copy/paste error from `session` and not this while file can go away.

jrjohnson (changes requested):
One comment and a minor change, otherwise LGTM.

addon/serializers/ilm-session.js
stopfstedt: nice find. that allows us to remove the ilm-specific serializer altogether and use the default one instead  :heavy_check_mark:

addon/components/session-overview-ilm-duedate.js
```diff
@@ -0,0 +1,55 @@
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { validatable, NotBlank } from 'ilios-common/decorators/validation';
+import { dropTask } from 'ember-concurrency';
+import moment from 'moment';
+
+@validatable
+export default class SessionOverviewIlmDuedateComponent extends Component {
+  @NotBlank() @tracked dueDate = null;
```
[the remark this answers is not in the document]
stopfstedt: ok, changed. we're guard-railing for the presence of `@ilmSession` before rendering the date- and time-controls in the component's template already, so no need to do the same in `SessionOverview` again.

jrjohnson (approved):

PR #2143, opened 2021-05-21
rework Length validation decorator to work on minimum-only input.

stopfstedt:
needed in frontend for validating user passwords since they don't have an upper boundary for length.

this turned into a more substantial rewrite, of note are:
- checking the length of the constraints array turned out to be pointless. it is always 2. instead, check if the values in that array are `undefined` to verify that actual parameters were provided.
- use the [`isBlank`](https://api.emberjs.com/ember/release/functions/@ember%2Futils/isBlank) utility function to check whether the given value is empty or not.
- input processing in the `validate()` and `defaultMessage()` methods was slightly different and have been realigned.

### DIFF
--- a/addon/decorators/validation/length.js
+++ b/addon/decorators/validation/length.js
@@ -1,5 +1,6 @@
 import { registerDecorator } from 'class-validator';
 import { getOwner } from '@ember/application';
+import { isBlank } from '@ember/utils';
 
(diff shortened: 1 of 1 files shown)


stopfstedt:
refs https://github.com/ilios/frontend/pull/6069

jrjohnson (approved):
I accept this.

PR #2158, opened 2021-05-28
Give the events grid some columns

### LINKED ISSUE #2157: Custom Non-5pm  ILM Events (and others) Spilling over to Other Days in Mobile View on Calendar
This has been reported using Android phones with Firefox and using Chrome in small window size on a Mac. I'm going to paste the screen shots in for reference. 

image

Those purple bars that are being pointed out on the calendar are ILM's that can only ever be due on one date and time combination. There is no situation possible where they could spill out over to other additional days.

The following / upcoming screen shot shows that the ILM's do not always extend into other days. This happens when there is already adjacent information on the calendar. 

image

jrjohnson:
We use span 50 and fractions of that to place events on the calendar
along side other events at the same time. This mostly worked, but
without specifying the number of columns in the grid sometimes it
overflows on mobile. This fix uses 50 evenly spaced columns to fill out
each day of the calendar.

Fixes #2157

### DIFF
--- a/app/styles/ilios-common/components/daily-calendar.scss
+++ b/app/styles/ilios-common/components/daily-calendar.scss
@@ -57,6 +57,7 @@
       display: grid;
       grid-column: 2;
       grid-row: 1 / -1;
+      grid-template-columns: repeat(50, 1fr);
       grid-template-rows: repeat($minute-rows, $hour-height);
       border: 1px solid $grid-border-color;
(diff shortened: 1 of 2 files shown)


dartajax (approved):
click tested and checked different calendar views - no issues to report at this time

dartajax:
fixes https://github.com/ilios/ilios/issues/3466

Issue #2171, opened 2021-06-02
ILM Callout On Calendar Needs Updating - to include Time of Day


dartajax:
Here's Week at a Glance showing the now customized ILM Due dates which include time of day.

![image](https://user-images.githubusercontent.com/7493331/120535739-31083980-c398-11eb-87ad-e3c79ba0a298.png)

Here is an example of a hover over on the Arteries and Arterioles due on Monday, May 31 at 8:55AM. The user sees this on the calendar - showing the callout as well.

![image](https://user-images.githubusercontent.com/7493331/120536001-7d537980-c398-11eb-90ee-7abfdf3213ef.png)

The callout should read "ILM - Due by 8:55 AM" rather than "ILM - Due this day". The user can already what day it is to which the calendar refers. This would be easier on students since the time of day is key information and not extremely easy to read in all calendar views without the assistance of this callout.

PR #2172, opened 2021-06-02
show start time for ILMs in calendar event tooltip.

stopfstedt:
image

 fixes https://github.com/ilios/common/issues/2171


dartajax (changes requested):
Can you change this ...?

`contents = contents + `ILM - ${startTime} - ${endTime}<br />${name}`;`

into ...

`contents = contents + `ILM Due by ${startTime} <br />${name}`;`

The reason for this that the 15 minute business in this context is irrelevant. It's due at the start time not the end.

dartajax (approved):
I see so it's in the translations where Dr. Finicky over here wants a lower case as in `Due by` - if it's translations, I'm good with this change.

PR #2175, opened 2021-06-03
preserve sort order of parent course objectives when managing session objectives

### LINKED ISSUE #2173: Course Objective Sort Order Needs To Persist in Parent Selection Process
Course objectives can be sorted at the Courses and Sessions level. This sort order should also be applied when attaching course objectives to session objectives. Currently this is not happening. Screen shots below ...

Here is a screen shot showing the adjusted sort order at the course level ...
image

Now when I want to add a parent (course) objective to a session objective here is what happens ...
image

The newest course objective appears at the bottom even though it was selected to appear first in the sort order. Sort order should apply here as well.

image

stopfstedt:
fixes #2173


dartajax (approved):
looks good to me

jrjohnson (approved):

PR #2179, opened 2021-06-04
Automerge some PRs

jrjohnson:
Using the github provided dependabot metadata action we can now
introspect pull requests to determine if they meet our criteria for
auto-merging. For now I've enable only those that where test coverage is
assured or that are part of the build.


stopfstedt (approved):
LGTM

PR #2195, opened 2021-06-09
always allow toggling of session/course leadership expanded/collapsed view

### LINKED ISSUE #2188: Missing Screen Element on Session Leadership - Expand / Contract
image

For reference, look at Objectives as shown below.

before ...

image

after ...

image

This same behavior should be available for Session Leadership.

stopfstedt:
the `isCollapsed` getters did not correctly recompute on leadership changes, among other things.
i threw this logic and other related state-management code out altogether in favor of a a less convoluted solution that _always_ let's the user expand/collapse the leadership components 

fixes #2188

### DIFF
--- a/addon/components/course-editing.hbs
+++ b/addon/components/course-editing.hbs
@@ -1,10 +1,5 @@
 <div>
-  {{#if
-    (or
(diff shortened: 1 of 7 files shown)


dartajax (approved):
works for me - DartAjax.approved=True

Issue #2203, opened 2021-06-11
Route User To Correct Week in All Weeks


dartajax:
I am not sure if this is possible but I believe it may be worth discussion at least. 

This is in All Weeks.

https://ilios3-demo.ucsf.edu/weeklyevents?expanded=38-41

In the example above, it is obvious that weeks 38-41 are expanded. I would like to see if it might be possible to send someone a link that is anchored to a particular expanded week.

crummy examples below ...

https://ilios3-demo.ucsf.edu/weeklyevents?expanded=38-41#selected=41

or ...

https://ilios3-demo.ucsf.edu/weeklyevents?expanded=40#selected=True

PR #2208, opened 2021-06-16
scroll the current week into view.

stopfstedt:
fixes https://github.com/ilios/common/issues/2203

this will pass the current week from the Dashboard to All Weeks when clicking the "All Weeks" link, and the scrolls down to the given week.

### DIFF
--- a/addon/components/dashboard-week.hbs
+++ b/addon/components/dashboard-week.hbs
@@ -4,7 +4,7 @@
 >
   <div class="weeklylink" data-test-weekly-link>
     {{t "general.view"}}:
-    <LinkTo @route="weeklyevents" @query={{hash expanded=this.expanded}}>
+    <LinkTo @route="weeklyevents" @query={{hash expanded=this.expanded week=this.week}}>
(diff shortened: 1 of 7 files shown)


dartajax (commented):
This is an improvement for sure.

stopfstedt:
> Given the title of the ticket that I entered, I approve this. There is another aspect to this that I'd like to explore as well. Users would have to manually change the URL to send an instructor to a specific week's schedule that is not the current week. The scrolling in position is excellent though. The week doesn't get updated in the URL -- for example ...
> 
> https://deploy-preview-2208--ilios-common.netlify.app/weeklyevents?expanded=29&week=24
> 
> I expanded week #29 because that is the one I want to send to an instructor - not week 24. Maybe this is not quite possible since the workaround would be to manually change the URL to ...

Good catch. The `week` parameter in the URL should now change as the user expands/collapses weeks. please give it another try.

dartajax (approved):
ein besserung - looks good

jrjohnson (approved):

PR #2210, opened 2021-06-16
adds session type to calendar event tooltip.

### LINKED ISSUE #2209: Add Session Type to Calendar Event Callout
image

The session above is a Lecture. This is shown after the user clicks on the event and reviews the event detail. It is true we have the functionality available to color code the appearance of session types on the calendar. This is not sufficient for many reasons - including accessibility standards. It has been suggested that we add session type to the calendar callout. I concur with this logic as long as it is possible spatially.

image

stopfstedt:
fixes #2209

### DIFF
--- a/addon/utils/calendar-event-tooltip.js
+++ b/addon/utils/calendar-event-tooltip.js
@@ -37,12 +37,24 @@ export default function calendarEventTooltip(event, intl, timeFormat) {
     return contents;
   };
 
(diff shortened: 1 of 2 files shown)


dartajax (approved):
I'd be honored to merge this after code review from @jrjohnson

jrjohnson (approved):

PR #2248, opened 2021-07-12
show timing window on blanked materials as well.

### LINKED ISSUE #1904: End Time Needs to Be Included in Event Detail

stopfstedt:
fixes #1904


jrjohnson (approved):

dartajax:
I have some concerns about the red font when the LM is no longer available - maybe grey or less aggressive color would be better - gonna merge though since that is a new ticket.

PR #2249, opened 2021-07-13
reorganize event details page

### LINKED ISSUE #2244: revamp event details page
- [x] move from two to one-column layout
- [x] render objectives as full markup
- [x] add collapser to session objectives, default is expanded
- [x] add collapser to course LMs, default is collapsed
- [x] add collapser to course objectives, default is collapsed
- [x] keep competency switcher where it is, it should be non-interactable/super-greyed-out when the objective section is collapsed

stopfstedt:
fixes #2244

### DIFF
--- a/addon/components/single-event-objective-list.hbs
+++ b/addon/components/single-event-objective-list.hbs
@@ -43,9 +43,9 @@
         <li>
           {{! template-lint-disable no-triple-curlies }}
           {{{objective.title}}}
-          <span class="details">
+          <div class="details">
             {{objective.domain}}
-          </span>
+          </div>
         </li>
       {{else}}
         <li class="no-content">
--- a/addon/components/single-event-objective-list.hbs
+++ b/addon/components/single-event-objective-list.hbs
@@ -36,7 +36,7 @@
               <ul>
                 {{#each domain.objectives as |title|}}
                   {{! template-lint-disable no-triple-curlies }}
-                  <li>
+                  <li class="objective">
                     {{{title}}}
                   </li>
                 {{/each}}
@@ -52,7 +52,7 @@
     {{else}}
       <ul class="list-in-order">
         {{#each @objectives as |objective|}}
-          <li>
+          <li class="objective">
             {{! template-lint-disable no-triple-curlies }}
             {{{objective.title}}}
             <div class="details">
--- a/addon/components/single-event-objective-list.hbs
+++ b/addon/components/single-event-objective-list.hbs
@@ -13,8 +13,9 @@
     </button>
     {{#if this.showDisplayModeToggle}}
       <button
-        class={{if this.groupByCompetencies "active"}}
+        class={{unless this.isExpanded "disabled" (if this.groupByCompetencies "active")}}
         type="button"
+        disabled={{not this.isExpanded}}
         {{on "click" (set this.groupByCompetencies (not this.groupByCompetencies))}}
       >
         {{#if this.groupByCompetencies}}
--- a/addon/components/single-event-objective-list.hbs
+++ b/addon/components/single-event-objective-list.hbs
@@ -8,6 +8,7 @@
       class={{if this.isExpanded "expanded" "expandable"}}
       type="button"
       {{on "click" (set this.isExpanded (not this.isExpanded))}}
+      data-test-expand-collapse
     >
       {{@title}}
     </button>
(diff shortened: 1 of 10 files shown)


dartajax (changes requested):
currently on demo3 ...

image

currently - same event in the netlify PR ...

image

There are some spacing and formatting issues here.

stopfstedt:
> There are some spacing and formatting issues here.

those formatting and spacing issues are introduced by the markup in the rendered contents, and not by the components rendering them. 

not sure if and how we should address this.

image

dartajax (approved):

addon/components/single-event-objective-list.hbs
```diff
@@ -1,11 +1,24 @@
-<div class="single-event-objective-list">
-  <h2>
-    {{@title}}
+<div
+  class="single-event-objective-list"
+  data-test-single-event-objective-list
+  {{did-insert (set this.isExpanded @isExpandedByDefault)}}
```
jrjohnson, line 4: I like this, but wonder if the constructor is a better place to set defaults like this from arguments. Not sure. Just wondering.

addon/components/single-event-objective-list.hbs
```diff
@@ -1,11 +1,24 @@
...
+      type="button"
+      {{on "click" (set this.isExpanded (not this.isExpanded))}}
+      data-test-expand-collapse
+    >
+      {{@title}}
+    </button>
     {{#if this.showDisplayModeToggle}}
       <button
-        class={{if this.groupByCompetencies "active"}}
+        class={{unless this.isExpanded "disabled" (if this.groupByCompetencies "active")}}
```
jrjohnson, line 17: I'm having trouble reading this one. Is it two expressions or one?
```suggestion
        class="{{unless this.isExpanded "disabled"}} {{if this.groupByCompetencies "active"}}"
```

addon/components/single-event-objective-list.hbs
```diff
@@ -15,42 +28,45 @@
...
+  {{#if this.isExpanded}}
+    {{#if this.groupByCompetencies}}
+      {{#if this.domains.length}}
+        <ul class="tree" data-test-tree>
+          {{#each this.domains as |domain|}}
+            <li data-test-domain>
+              <span data-test-domain-title>{{domain.title}}</span>
+              <ul>
+                {{#each domain.objectives as |title|}}
+                  {{! template-lint-disable no-triple-curlies }}
```
jrjohnson, line 40: I'm wondering in light of https://github.com/ilios/ilios/issues/3571 if we should create an ObjectiveTitle component that handles this and consolidates the no-tripple-currlies as well.

addon/components/single-event-objective-list.js
```diff
@@ -3,6 +3,7 @@ import { tracked } from '@glimmer/tracking';
 
 export default class SingleEventObjectiveList extends Component {
   @tracked groupByCompetencies = true;
+  @tracked isExpanded = false;
```
jrjohnson, line 6: I *think* this would work:
```suggestion
  @tracked isExpanded = this.args.isExpanded ?? false;
```

jrjohnson (changes requested):
Regarding @dartajax's comment I think some style could be applied here to any child div or p etc to zero out the margin and padding. Couple minor tweaks. Looks great!

addon/components/single-event-objective-list.hbs
stopfstedt: actually, that's one expression. i merged your suggestion too soon. undoing it...

stopfstedt: yes, so it's either disabled, or active. but cannot be both, since they both apply background-color styles to this button.

addon/components/single-event-objective-list.js
stopfstedt: :heavy_check_mark: went with something like that.

addon/components/single-event-objective-list.hbs
stopfstedt: :heavy_check_mark:

addon/components/single-event-objective-list.hbs
jrjohnson: Ah, I see it now. If the first is false then the second gets read and if that is true show up. Ok.

addon/components/single-event-objective-list.hbs
stopfstedt: maybe. seems a bit excessive to create a component wrapper just to render triple curlies. but maybe i'm not understanding the full scope of the issue here.

stopfstedt: follow-up item for sure.

stopfstedt:
@jrjohnson this is ready for re-review.

jrjohnson (approved):

PR #2262, opened 2021-07-28
Fix capitalization of header in test

jrjohnson:
This aligns with reality.

This fixes issues with out tests, but I'm not sure why it is necessary and what dependency update is making it necessary. I'm going to leave this in draft while I figure that out, otherwise we might have to repeat this exercise in a few days.

Ugh... and of course now the original test is failing. Something annoying here... investigating....

OK, cause is `pretender` and I've upgraded it here as well.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3423,9 +3423,9 @@
       }
     },
     "@ember/test-helpers": {
-      "version": "2.2.8",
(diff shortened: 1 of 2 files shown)


stopfstedt (approved):
nice sleuthing and thankfully easy fix for it. LGTM.

PR #2269, opened 2021-08-02
Add attributes to UserNameInfo component

jrjohnson:
This allows passing id in particular so the component can be used as an
arial-labeledby property.


stopfstedt (approved):
LGTM

dartajax (approved):

Issue #2226, opened 2021-06-25
Add Breadcrumbs to Visualizations


dartajax:
This is sort of an enhancement request and will be tagged as such. The issue from my perspective is that it would help with the navigation in Visualizations (Course level).

![image](https://user-images.githubusercontent.com/7493331/123466959-cd42ec00-d5a4-11eb-898a-8d35850f4a61.png)

![image](https://user-images.githubusercontent.com/7493331/123467283-2f035600-d5a5-11eb-83af-c5951ef04f1d.png)

![image](https://user-images.githubusercontent.com/7493331/123467667-9f11dc00-d5a5-11eb-9fba-2b1f3c93f460.png)

Maybe include the page Title as well so it looks like the following for nav purposes - the graph from the previous screen shot would be below all of this.

![image](https://user-images.githubusercontent.com/7493331/123469000-58bd7c80-d5a7-11eb-8b18-f59b20d1328f.png)

PR #2295, opened 2021-08-12
Add breadcrumbs to course visualizations

stopfstedt:
fixes https://github.com/ilios/common/issues/2226


dartajax (changes requested):
any chance we can have `Other Visualizations` become `Visualizations` instead?

image

It's in the same context as `Learner Groups` now which does not say `Other Learner Groups` - dig?

stopfstedt:
> any chance we can have `Other Visualizations` become `Visualizations` instead?
you got it :+1:

dartajax (approved):

jrjohnson (approved):

PR #2306, opened 2021-08-19
corrects display of "no materials available" message when applicable.

### LINKED ISSUE #2305: Materials View Not Completing Load Cycle When No Course LM's Exist
Try this one on to see what I'm on about.

https://ilios3-demo.ucsf.edu/courses/1875/materials

image

This occurs when a Course has no learning materials at the course level. I will test if the same is true for session materials. For now, it seems evident that if there are no course materials, the loading spinner never completes its job.

stopfstedt:
fixes #2305

### DIFF
--- a/addon/components/course-materials.hbs
+++ b/addon/components/course-materials.hbs
@@ -49,7 +49,7 @@
         </tr>
       </thead>
       <tbody>
-        {{#unless this.courseLearningMaterialObjects}}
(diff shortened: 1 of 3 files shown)


dartajax (approved):
looks good to me

dartajax:
This also fixes the situation when there are no session learning materials - so ... great.

PR #2317, opened 2021-08-30
fixes various issues with PCRS and mirage.

stopfstedt:
typo fix, and adds kludge for retrieving all PCRS.


jrjohnson (approved):
LG in so far as it looks normal. I accept this.

PR #2363, opened 2021-09-30
Add link to session to single event details

### LINKED ISSUE #2362: Provide Link to Course/Session Detail for Non-Student Ilios Users in Event Detail
The link for these users would be to ...

[Ilios_URL]/courses/[course_id]/sessions/[session_id]

example URL ...

https://ilios3-demo.ucsf.edu/courses/1912/sessions/62513

These non-student Ilios users have access to the Courses and Sessions page and we can have it in read-only mode thanks to the permissions matrix that has been implemented.

image

jrjohnson:
This allows those with elevated privileges to navigate directly to the
session for an event from the single event details page.

Fixes #2362


dartajax (approved):
It looks fine to me.

PR #2370, opened 2021-10-05
Don't copy prerequisites when copying session

### LINKED ISSUE #2368: Copying Session is a Problem when there are Prerequisites
In https://github.com/ilios/common/pull/470 we added prerequisites to session copy. However this doesn't actually work and doesn't make sense. A session can only have one postrequisite so copying the prerequisites for a session either destroys the original links from the session to be copied or does not have any effect. This is further confused by the copy working in the UI when it probably didn't actually change the data in the API. 

We probably just need to drop copying in this case, but I wanted to open for discussion first.

jrjohnson:
This won't work because on the other side of the relationship a session
can't have more than one postrequisite.

Fixes #2368


dartajax (approved):
Code looks good and it works well - gonna merge this now.

PR #2431, opened 2021-10-31
Fix a11y on multiday calendar events

jrjohnson:
Switched to a real button inside the list instead of a list item
pretending to be a button.


stopfstedt (approved):
LGTM

PR #2434, opened 2021-11-02
Implement a type scale system for font-size and line-height

jrjohnson:
By replacing calls to font-size with a mixing `@include font-size('large');` we can standardize the type sizes across the site and also ensure that proper line-height is applied at each step.

Wip:
- [x] needs extensive visual review
- [x] Check with a11y experts if this meets the needs of AT, we're using a dynamic line height here which doesn't make Siteimprove happy, but should be good for real users.
- [x] Find a fix for extra `p` tags in session descriptions (and anywhere else)

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,5 +1,6 @@
 {
   "extends": "stylelint-config-recommended-scss",
   "rules": {
+    "property-disallowed-list": ["font-size", "line-height"],
   }
 }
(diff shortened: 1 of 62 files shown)


stopfstedt (approved):
code reviewed and visually spot-checked. LGTM.

dartajax:
From my testing earlier in the day, I'm fine with it too.

dartajax:
I will merge this one in the morning to give @saschaben a chance to review it too. I'm fine with it as I have said.

PR #2438, opened 2021-11-04
Remove italics from LM note

jrjohnson:
Having an entire paragraph in italics isn't accessible.


dartajax (approved):
I tested this - looking to make sure no italics appear on LM notes - none did. Looks fine.

PR #2439, opened 2021-11-04
Rework dashboard headings

jrjohnson:
Align the heading number and position so headings fall in order. Also
removed headings from several places where we were using them mainly for
styling and not content.


stopfstedt (approved):
looks much better.

PR #2444, opened 2021-11-08
Update ember-drag-drop

jrjohnson:
This beta version fixes deprecations we need to clear.


dartajax (approved):
Continuing to click test for a bit more but it looks good to me - tested Course Objectives, Session Objectives, Course LM's - will still test Session LM's and more.

stopfstedt (approved):
click-tested. LGTM.

PR #2446, opened 2021-11-08
Label agenda links for a11y

jrjohnson:
I've removed the link from the room and labeled the link in the first
column with both the name of the event and the date. This ensures that we
don't present a hundred links with the exact same name in a confusing
way.


stopfstedt (approved):
LGTM

PR #2455, opened 2021-11-14
Extract routes into an export

jrjohnson:
This will allow importing in our apps and remove duplication as well as
ensuring that changes sync correctly.


stopfstedt (approved):
LGTM

PR #2456, opened 2021-11-14
Add explicit model hooks to course routes

jrjohnson:
We were relying on the automatic model hook here, but that will soon be
deprecated.


stopfstedt (approved):
LGTM

PR #2457, opened 2021-11-14
Use parts of translation again

jrjohnson:
Reverts [d9cef043b17b045664fec3e1025fdf8920e84d8e] as this is now fixed
upstream and will no longer fail on the initial undefined value.


stopfstedt (approved):
nice.

PR #2466, opened 2021-11-18
Scale fonts in a smaller range

jrjohnson:
Instead of scaling from phone to desktop this keeps us at 16px until we
gt up to laptop size and then scales up to the 20x for larger screens.


dartajax (approved):
I like it.  Testing on laptop.

stopfstedt (approved):

PR #2467, opened 2021-11-18
Octaneify Instructor Group Model

jrjohnson:


addon/models/user.js
```diff
@@ -1,202 +1,263 @@
...
+      `user.performsNonLearnerFunction called, should be read from current-user session.`,
+      false,
+      {
+        id: 'common.user-performs-non-learner-function',
+        for: 'ilios-common',
+        until: '65',
+        since: '62.0.0',
       }
+    );
+    return Boolean(
```
stopfstedt, line 227: i don't think this needs to be wrapped in a Boolean constructor. the result of the disjunction is always a boolean.

addon/models/user.js
```diff
@@ -206,171 +267,135 @@ export default Model.extend({
...
+      displayName === this.fullNameFromFirstMiddleInitialLastName.trim().toLowerCase()
+    );
+  }
+
+  @use _instructedLearnerGroupOfferings = new ResolveFlatMapBy(() => [
+    this._instructedLearnerGroups,
+    'offerings',
+  ]);
+
+  get _instructedOfferings() {
```
stopfstedt, line 298: private getter?

stopfstedt (approved):
clean, or at least as clean as we can be with this new pattern. LGTM,

PR #2470, opened 2021-11-19
octanifies Offering model.

stopfstedt:


jrjohnson (approved):
Sweet! I maybe like these patterns again now.

dartajax (approved):
'ember-could-get-used-to-this'? it seems to work - clicked through offering editor - no issues

PR #2471, opened 2021-11-19
Resolve objectives in course publication check

jrjohnson:
We can also use the length value of the relationship to check if there
are parents which engages auto tracking.


stopfstedt (approved):

dartajax (approved):
I did some publishing tests and some clickaround - looks good

PR #2480, opened 2021-11-24
Fix publish all / schedule all buttons

### LINKED ISSUE #2478: TypeError: Cannot set property sessionsToOverride of #<d> which has only a getter
Sentry Issue: [FRONTEND-1J2](https://sentry.io/organizations/ilios/issues/2811380786/?referrer=github_integration)

```
TypeError: Cannot set property sessionsToOverride of #<d> which has only a getter
  at _default (addon-tree-output/ilios-common/components/publish-all-sessions.js:509:1)
  at HTMLButtonElement.n (addon-tree-output/moment/lib.js:22:1)
```

jrjohnson:
The not not array wasn't working, I think this is more explicit and
readable, plus it actually works.

Fixes #2478

### DIFF
--- a/addon/components/publish-all-sessions.hbs
+++ b/addon/components/publish-all-sessions.hbs
@@ -194,7 +194,7 @@
       {{#if this.overridableSessions.length}}
         <button
           type="button"
-          disabled={{this.allSessionsAsIs}}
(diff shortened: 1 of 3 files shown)


stopfstedt (approved):
the new getter names are :fire: , nice improvement.

PR #2490, opened 2021-11-29
converts School model to class syntax.

stopfstedt:
removed unused methods while at it.

### DIFF
--- a/addon/models/school.js
+++ b/addon/models/school.js
@@ -1,75 +1,57 @@
 import Model, { hasMany, belongsTo, attr } from '@ember-data/model';
-import { computed } from '@ember/object';
 import { isEmpty } from '@ember/utils';
-import RSVP, { filter } from 'rsvp';
-const { map } = RSVP;
+import { deprecate } from '@ember/debug';
 
-export default Model.extend({
-  title: attr('string'),
-  templatePrefix: attr('string'),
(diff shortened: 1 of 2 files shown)


addon/models/school.js
```diff
@@ -1,75 +1,61 @@
...
-   * @return {Promise.<Array>}
-   */
-  async getProgramYearsForYear(year) {
-    const programs = await this.programs;
-    const rhett = await map(programs.mapBy('programYears'), (programYears) => {
-      return programYears.filter((programYear) => {
-        return parseInt(programYear.get('startYear'), 10) === year;
-      });
+  get cohorts() {
+    deprecate(`school.cohorts is called, use getCohorts() instead.`, false, {
```
jrjohnson, line 36: I think we should encourage ourselves to just get the cohorts by loading everything, pulling the cohorts off the store guarantees a network request even if the data is already loaded. My preference would be for this getter to return the `store.query` promise from `getCohorts` and for us to stop using it ASAP.

jrjohnson (changes requested):

addon/models/school.js
stopfstedt: updated :heavy_check_mark:

jrjohnson (approved):

PR #2513, opened 2021-12-20
dedupes associated learner groups display

### LINKED ISSUE #2512: Associated Groups - Session - Eliminate Duplicated Learner Groups
This example is utilizing recurring events - there are six event these learner groups are scheduled to attend.

The learner group name should not be repeated eight times and this is loosely related to ilios/ilios#3790 in the sense that sorting should be applied here as well. Only show each associated learner group once. 

Screen Shot 2021-12-14 at 3 21 03 PM

stopfstedt:
fixes #2512


jrjohnson (approved):

dartajax (approved):
click tested - looks great

PR #2522, opened 2021-12-21
Silencing a couple of very loud deprecations

jrjohnson:
We'll need to solve these, but at least our logs are more readable when
they're quiet.

### DIFF
--- a/tests/dummy/config/deprecation-workflow.js
+++ b/tests/dummy/config/deprecation-workflow.js
@@ -5,5 +5,15 @@ window.deprecationWorkflow.config = {
   workflow: [
     { handler: 'silence', matchId: 'manager-capabilities.modifiers-3-13' }, //https://github.com/emberjs/ember-render-modifiers/issues/32
     { handler: 'silence', matchId: 'ember-global' },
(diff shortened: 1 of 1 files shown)


tests/dummy/config/deprecation-workflow.js
```diff
@@ -5,5 +5,10 @@ window.deprecationWorkflow.config = {
   workflow: [
     { handler: 'silence', matchId: 'manager-capabilities.modifiers-3-13' }, //https://github.com/emberjs/ember-render-modifiers/issues/32
     { handler: 'silence', matchId: 'ember-global' },
+    { handler: 'silence', matchId: 'common.async-computed' },
+    { handler: 'silence', matchId: 'common.async-computed' },
```
stopfstedt, line 9: duplicate entry, please remove.

stopfstedt (changes requested):

tests/dummy/config/deprecation-workflow.js
jrjohnson: ✅ plus I added a few that were missing.

stopfstedt (approved):

PR #2523, opened 2021-12-21
Inject a missing store

jrjohnson:
This is needed here and we're relying on default injection which is
going away in ember-data v4.


stopfstedt (approved):
nice find.

Issue #2524, opened 2021-12-16
Month View Not Showing Multi-Day Events on Calendar Correctly


dartajax:
I am logged in as Nicholas Gutierrez - school admin and course director.

![image](https://user-images.githubusercontent.com/7493331/146284274-5c6a46c3-bdf1-4295-a9b1-d7a578b2cd01.png)

Here is the event (only the first one) on the Calendar - Each day in the Month View between Xmas eve and Jan 2, 2022 should have one of these offerings displayed on that date. 

How it is ..

<img width="1260" alt="Screen Shot 2021-12-15 at 4 09 51 PM" src="https://user-images.githubusercontent.com/7493331/146284404-f6471dbe-7c9c-40aa-bb61-d2b94257ffdd.png">

How it should be -- this is an example from the User Guide ...

<img width="801" alt="Screen Shot 2021-12-15 at 4 10 46 PM" src="https://user-images.githubusercontent.com/7493331/146284490-2b717e74-e0ec-415d-a07b-69de3b89377f.png">

I left the old screen shot in the User Guide for now - since this bug made it not correct to update it.

mhgh22:
Month view currently showing multi-day events. -MNS @dartajax 

![Screenshot 2022-04-21 at 09-42-19 Ilios Dashboard](https://user-images.githubusercontent.com/103135575/164510631-2fd0f5ae-27ce-41c9-b0af-5a1e5d5bfdb4.png)

PR #2525, opened 2021-12-21
adds multiday events display to monthly calendar view.

stopfstedt:
this is the quick- and easy fix getting multi-day events on the display, it's also consistent on how it's handled on other calendar displays (month and day view).

refs #2524

todo:
- [x] update test coverage

### DIFF
--- a/addon/components/ilios-calendar-month.js
+++ b/addon/components/ilios-calendar-month.js
@@ -1,5 +1,5 @@
 import Component from '@glimmer/component';
-import moment from 'moment';
+import { DateTime } from 'luxon';
 import { action } from '@ember/object';
 
 export default class IliosCalendarMonthComponent extends Component {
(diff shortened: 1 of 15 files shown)


dartajax (approved):
month view - nice

addon/components/ilios-calendar-month.js
```diff
@@ -23,6 +24,18 @@ export default class IliosCalendarMonthComponent extends Component {
     });
   }
 
+  get singleDayEvents() {
+    return this.nonIlmPreWorkEvents.filter((event) =>
+      moment(event.startDate).isSame(moment(event.endDate), 'day')
```
jrjohnson, line 29: Can you swap these with [luxon](https://moment.github.io/luxon/#/math?id=comparing-datetimes)? I'd like to push to get rid of moment at some point this year. Example of importing and using it is in [session-object](https://github.com/ilios/common/blob/96e9981a1a9a7d6d80a13d1890fc64e9d0f5ec40/addon/classes/session-object.js#L7)

jrjohnson (changes requested):

addon/components/ilios-calendar-month.js
stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #2528, opened 2021-12-21
corrects multi-day output by converting method into getter.

### LINKED ISSUE #2527: Day View Not Showing Multi-Day Events on Calendar Correctly
This is related to https://github.com/ilios/common/issues/2524 in the sense that multi-day events don't show up correctly on the Calendar.

Here is how the event was set up in Courses and Sessions ...

Screen Shot 2021-12-15 at 4 15 37 PM

This event goes from 12/23 to 1/2 8am.

I am logged in as the Course Director it should be pointed out. I will test Student View as well but wanted to document this anyway.

Here is the day view of Dec. 24th, which should be indicated as having the multi-day holiday event. I have scrolled to the bottom to show it does not appear anywhere.

stopfstedt:
fixes #2527

### DIFF
--- a/addon/components/ilios-calendar-day.js
+++ b/addon/components/ilios-calendar-day.js
@@ -40,7 +40,7 @@ export default class IliosCalendarDayComponent extends Component {
       this.moment.moment(event.startDate).isSame(this.moment.moment(event.endDate), 'day')
     );
   }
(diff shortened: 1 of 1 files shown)


jrjohnson (approved):
Whoops! Awesome fix 😁

PR #2545, opened 2022-01-05
Replace usage of is helper in EPO tests

jrjohnson:
Ember page object deprecated the "is" helper. In our case, finding
focus, we need a custom helper to replace it.


stopfstedt (approved):
nice!

PR #2563, opened 2022-01-13
chore: downgrade colors to 1.4.0 in package-lock.json

Trott:


Trott:
Not sure what the "Floating Dependencies" failure is about. Is that something I need to worry about?

jrjohnson:
That's with no lock file.

Thanks for this, but I'm going to close in favor of #2562 as we need to fix the floating error anyway. I'll also dig into why the action is failing I'm guessing it's published with something in it's own lock file...

jrjohnson:
After fiddling with it and hitting other problems I changed my mind and could love to merge this! Any chance you can bring it back @Trott?

PR #2565, opened 2022-01-14
Remove moment shim addon

jrjohnson:
Going to try importing everything ourselves and not using this anymore
if possible.

Weirdly this seems to work with no changes needed at all and works in frontend as well. I thinking we make this change and mixup any issues if they pop up. It's very likely this is no longer needed as ember-auto-import can no find everything it needs directly.


jrjohnson:
Closing in favor of #2590

PR #2574, opened 2022-01-14
(temporarily) disable tests against floating dependencies.

stopfstedt:


stopfstedt:
per discussion with @jrjohnson, we don't want to disable the Floating Dependencies step at this point.

PR #2576, opened 2022-01-15
[breaking] Fix deprecated mirage imports

jrjohnson:
These aren't going to be provided by mirage anymore, we can get them
from ember/string though, no problem.


stopfstedt (approved):
LGTM

PR #2588, opened 2022-01-21
[breaking] Move to new mirage config style

jrjohnson:
The old route only style has been deprecated and we need to make a few
adjustments to keep moving forward.


stopfstedt (approved):
super.

PR #2604, opened 2022-01-25
Bump Node to v16 and pin @embroider/macros

jrjohnson:
This should unblock us while we sort out getting macros updated everywhere else.


stopfstedt (approved):
click tested and reviewed. LGTM

PR #2605, opened 2022-01-26
Replace clipboard addon with component

jrjohnson:
Since we're only copying text that we know about and we don't need to
read it from another element it looks like all of our targeted browsers
support this simple method of getting the data onto the clipboard. Thus
allowing us to remove this seldom updated addon.

### DIFF
--- a/tests/integration/components/copy-button-test.js
+++ b/tests/integration/components/copy-button-test.js
@@ -1,0 +1,20 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | copy-button', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
+    await render(hbs`
+      <CopyButton>
+        template block text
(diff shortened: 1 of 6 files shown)


tests/integration/components/copy-button-test.js
```diff
@@ -0,0 +1,20 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | copy-button', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it renders', async function (assert) {
```
stopfstedt, line 9: would like a test case for the passed-in click-handler.

stopfstedt (changes requested):

tests/integration/components/copy-button-test.js
jrjohnson: Me too! Unfortunately if you try and write to the clipboard in a text context it dies because you don't have permission to do this. It's.... annoying.

stopfstedt: okely dokely.

stopfstedt (approved):

PR #2617, opened 2022-01-30
Get school from learner group better

jrjohnson:
This getter has been deprecated and it's not really suitable for instant
access anymore, instead we can get to school the old fashion way.


jrjohnson:
Going to self-merge to hopefully get frontend updated today.

PR #2626, opened 2022-02-03
Fix exports

jrjohnson:
Seems like the ember blueprints have change a few times here and we've
been left with a couple of places where we're exporting something that didn't
actually exist.


stopfstedt (approved):
LGTM.

PR #2628, opened 2022-02-04
Replace drag/drop sorting addon with our own implementation

jrjohnson:
This addon isn't updated anymore and the author doesn't recommend using it as drag/drop is better handled directly in the app. I'm not so sure about that as this was a *huge* pain to figure out, however I think it's working now and mostly OK. I couldn't duplicate the old interface exactly, hopefully this one is OK. Feedback on look/feel very much welcome.


dartajax (approved):
I tested dragging and dropping / re-ordering of learning materials and objectives - seems fine to me.

stopfstedt (approved):
i accept this.

PR #2631, opened 2022-02-07
Restore node 14.x support

jrjohnson:
Now that embroider is better up to date across our dependencies we no
longer need the override option in package.json and can relax our
required node version.


jrjohnson:
This can be merged with the missing test, once it is merged we'll change the requirements for what tests are required.

stopfstedt (approved):

PR #2646, opened 2022-02-09
Fortawesome rollback

stopfstedt:
the FortAwesome v6 upgrade does not work, icons we need are missing from the standard icon pack. 
So we must roll this back entirely.


stopfstedt:
refs https://github.com/ilios/common/pull/2643

jrjohnson (approved):

PR #2665, opened 2022-02-17
Some more dependency organization

jrjohnson:
This is the third go around with this, I'll get it right
eventually. Maybe.


jrjohnson:
Merging my own 🗑️ 🔥 here in hopes of getting it into the frontend and tested.

PR #2667, opened 2022-02-18
Back out of beta mirage and file upload addons

jrjohnson:
Handle the conflict with embroider macros using an override instead.


jrjohnson:
This doesn't work and will probably still break on the frontend. 😢

PR #2685, opened 2022-02-25
consolidate my-materials views

stopfstedt:
fixes https://github.com/ilios/frontend/issues/5248

fixes: ilios/common#2706


stopfstedt:
i'm going to stop here and solicit feedback before proceeding, since this is turning into a much more invasive refactoring than anticipated.

for next steps i'm considering breaking the `dashboard` route up into subroutes for the main four dashboard areas. reason being that i don't think the current dashboard route can handle any more query params, and i'll need to add back the params from the retired `my-materials` route (course filter, sorting options etc), as well add another flag to make the selected my-materials view (all materials vs next 60 days) sticky

@dartajax @saschaben please review and let me know if the UI changes is on track.


@jrjohnson please review and advise regarding the suggested changes to routing.

jrjohnson:
I was considering breaking this up for a11y reasons as well so we can clearly indicate that the page content has changed. It would let us drop some query params that are all jammed together now as well. So for that and a couple other reasons Im' 100% for breaking this up into several routes.

addon/components/dashboard/materials.js
```diff
@@ -0,0 +1,166 @@
...
+  @use _course = new AsyncProcess(() => [this.loadCourse.bind(this), this.args.courseIdFilter]);
+  @use _materials = new AsyncProcess(() => [
+    this.loadMaterials.bind(this),
+    this.args.showAllMaterials,
+  ]);
+
+  async loadMaterials() {
+    const user = await this.currentUser.getModel();
+    let url = `${this.iliosConfig.apiNameSpace}/usermaterials/${user.id}`;
+    if (!this.args.showAllMaterials) {
```
jrjohnson, line 29: I think maybe we should go ahead and load everything here and then filter it down based on `showAllMaterials`.

addon/controllers/dashboard.js
```diff
@@ -1,127 +1,3 @@
...
-  @action
-  clearFilters() {
-    this.cohorts = '';
-    this.courseLevels = '';
-    this.courses = '';
-    this.sessionTypes = '';
-    this.terms = '';
-  }
-}
+export default class DashboardController extends Controller {}
```
jrjohnson, line 3: This is the same as not having the file at all, so it can probably be removed.

addon/controllers/dashboard/activities.js
```diff
@@ -0,0 +1,3 @@
+import Controller from '@ember/controller';
+
+export default class DashboardActivitiesController extends Controller {}
```
jrjohnson, line 3: This is the same as not having the file at all, so it can probably be removed.

addon/controllers/dashboard/week.js
```diff
@@ -0,0 +1,3 @@
+import Controller from '@ember/controller';
+
+export default class DashboardWeekController extends Controller {}
```
jrjohnson, line 3: This is the same as not having the file at all, so it can probably be removed.

addon/routes/dashboard.js
```diff
@@ -0,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default class DashboardRoute extends Route {
+  beforeModel() {
+    this.transitionTo('dashboard.week');
```
jrjohnson, line 5: `transitionTo` on the route is deprecated. Need to import the router service see https://github.com/ilios/common/pull/2688/commits/0db3bea18c8428ee78418c9502b746dc706e291f

addon/routes/dashboard/activities.js
```diff
@@ -0,0 +1,3 @@
+import Route from '@ember/routing/route';
+
+export default class DashboardActivitiesRoute extends Route {}
```
jrjohnson, line 3: This is the same as not having the file at all, so it can probably be removed.

addon/routes/dashboard/calendar.js
```diff
@@ -0,0 +1,3 @@
+import Route from '@ember/routing/route';
+
+export default class DashboardCalendarRoute extends Route {}
```
jrjohnson, line 3: This is the same as not having the file at all, so it can probably be removed.

addon/routes/dashboard/index.js
```diff
@@ -0,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default class DashboardIndexRoute extends Route {
+  beforeModel() {
+    this.replaceWith('dashboard.week');
```
jrjohnson, line 5: deprecated see above.

addon/routes/dashboard/materials.js
```diff
@@ -0,0 +1,3 @@
+import Route from '@ember/routing/route';
+
+export default class DashboardMaterialsRoute extends Route {}
```
jrjohnson, line 3: This is the same as not having the file at all, so it can probably be removed.

addon/routes/dashboard/week.js
```diff
@@ -0,0 +1,3 @@
+import Route from '@ember/routing/route';
+
+export default class DashboardWeekRoute extends Route {}
```
jrjohnson, line 3: This is the same as not having the file at all, so it can probably be removed.

jrjohnson (changes requested):

addon/components/dashboard/materials.js
stopfstedt: impossible without changing what's being delivered on the frontend. the date range data points are not part of the materials payload.

addon/routes/dashboard/activities.js
stopfstedt: :heavy_check_mark:

addon/routes/dashboard/calendar.js
stopfstedt: :heavy_check_mark:

addon/controllers/dashboard/week.js
stopfstedt: :heavy_check_mark:

addon/controllers/dashboard/activities.js
stopfstedt: :heavy_check_mark:

addon/controllers/dashboard.js
stopfstedt: :heavy_check_mark:

addon/routes/dashboard/index.js
stopfstedt: :heavy_check_mark:

addon/routes/dashboard/materials.js
stopfstedt: :heavy_check_mark:

addon/routes/dashboard/week.js
stopfstedt: :heavy_check_mark:

addon/components/dashboard/materials.js
jrjohnson: Ah, ok. Yes. I remember you were saying this. Ok. Something to fixup on the API. 🆒

addon/routes/dashboard/index.js
```diff
@@ -0,0 +1,10 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class DashboardIndexRoute extends Route {
```
jrjohnson, line 4: I'm not sure this route is needed. Since `dashboard` redirects I don't think this will ever get called.

jrjohnson (approved):
One maybe thing. Otherwise I'm so happy about this change!

dartajax (approved):
looks great to me - tested only the Materials link from the Dashboard - not the one from the drop-down - I assume that has similar behavior - great stuff

stopfstedt:
> looks great to me - tested only the Materials link from the Dashboard - not the one from the drop-down - I assume that has similar behavior - great stuff

PR #2728, opened 2022-03-17
Track session sorting param

jrjohnson:
As this is consumed by a getter now and not by the template it doesn't
get autotracked.

Fixes ilios/ilios#3941


dartajax (approved):
I click tested this. Looks good to me.

stopfstedt (approved):

PR #2749, opened 2022-03-24
adds collapser to truncated text

stopfstedt:
fixes https://github.com/ilios/common/issues/2748

should also be tested with edit-in-place components that can receive "overlong" text, like the session description editor.


dartajax (approved):
DartAjax / Dart_500 approves. Looks great - love the collapser.

jrjohnson (approved):

Issue #2746, opened 2022-03-24
use page objects everwhere


stopfstedt:
companion ticket to https://github.com/ilios/frontend/issues/6581

```
$ egrep -lr '(assert.dom| find\(| findAll\(| click\()' tests | sort
tests/acceptance/course/printcourse-test.js
tests/acceptance/dashboard/calendar-test.js
tests/integration/components/back-link-test.js
tests/integration/components/course/loading-test.js
tests/integration/components/course/manage-objective-parents-item-test.js
tests/integration/components/course/objective-list-loading-test.js
tests/integration/components/course/rollover-date-picker-test.js
tests/integration/components/course-rollover-test.js
tests/integration/components/course-summary-header-test.js
tests/integration/components/daily-calendar-event-test.js
tests/integration/components/dashboard/agenda-test.js
tests/integration/components/dashboard/filter-tags-test.js
tests/integration/components/detail-cohort-list-test.js
tests/integration/components/detail-terms-list-item-test.js
tests/integration/components/detail-terms-list-test.js
tests/integration/components/editable-field-test.js
tests/integration/components/expand-collapse-button-test.js
tests/integration/components/html-editor-test.js
tests/integration/components/ilios-calendar-event-month-test.js
tests/integration/components/leadership-collapsed-test.js
tests/integration/components/leadership-list-test.js
tests/integration/components/leadership-manager-test.js
tests/integration/components/leadership-search-test.js
tests/integration/components/learning-materials-sort-manager-test.js
tests/integration/components/learning-material-uploader-test.js
tests/integration/components/lm-type-icon-test.js
tests/integration/components/loading-spinner-test.js
tests/integration/components/new-objective-test.js
tests/integration/components/new-offering-test.js
tests/integration/components/objective-sort-manager-test.js
tests/integration/components/offering-calendar-test.js
tests/integration/components/offering-url-display-test.js
tests/integration/components/progress-bar-test.js
tests/integration/components/pulse-loader-test.js
tests/integration/components/save-button-test.js
tests/integration/components/selectable-terms-list-item-test.js
tests/integration/components/selectable-terms-list-test.js
tests/integration/components/session-copy-test.js
tests/integration/components/session/manage-objective-parents-item-test.js
tests/integration/components/session/objective-list-loading-test.js
tests/integration/components/sessions-grid-last-updated-test.js
tests/integration/components/sessions-grid-loading-test.js
tests/integration/components/sessions-grid-offering-test.js
[...]

PR #2758, opened 2022-03-29
session description expands to HTML in WaaG events

### LINKED ISSUE #2757: Provide Better Formatting of Session Description in Week at a Glance
We will soon have collapsers/expanders on Week at a Glance and other places. This will make it so if truncated the full session description would be displayed after expanding the view. See the three views below for reference.

This is what the viewer sees in Week at a Glance ...

image

This is what the viewer sees in Event Detail ...

image

This is what the viewer sees in Courses and Sessions ...

image

The Week at a Glance view (the top one) does not format the text properly and makes it hard to read compared with the other two views that are rendering what was entered into the froala text boxes used in Ilios.

This last view shows the froala text box in edit mode for reference. 

Screen Shot 2022-03-29 at 1 09 46 PM

stopfstedt:
fixes #2757 

refs #2746


dartajax (approved):
looks great to me

Issue #2775, opened 2017-04-04
school picker on filters not sticky


saschaben:
when shifting between "show filters" and "hide filters" on the dashboard calendar, if a school other than the primary school has been selected, when the toggle is shifted the school will revert to the primary school. eg, if I am looking at SOP (as a SOM user) and I open/close the filters, the calendar will flip back from SOP to SOM.

mhgh22:
Able to duplicate issue after clicking on 'All Events" and following the original instruction. Issue is still present(Thanks Stefan). -Michael/Nanci/Stacey

PR #2776, opened 2022-04-07
prevent selected school from resetting when hiding filters.

stopfstedt:
fixes https://github.com/ilios/common/issues/2775


dartajax (approved):
wow - remove one line of code and bam - all good - works for me

Issue #2764, opened 2021-03-01
Formatting Suggestion - Vertical Spacing Improvement 


dartajax:
This is in the area where Terms are added to Session Objectives.

![image](https://user-images.githubusercontent.com/7493331/109565138-e1fbf100-7a96-11eb-8e35-bdbc4eee0c28.png)

The above would look better if ...

a. there was a little bit of vertical padding above the Select Vocabulary and "Start typing to filter list" fields
b. the same extra space underneath "UCSF 49 (Medicine) could be used for this. Well I'm sure it's "different space" but the screen shot below suggests the minor improvement I am suggesting.

![image](https://user-images.githubusercontent.com/7493331/109566115-45d2e980-7a98-11eb-94a1-e94fc1ce70d9.png)

PR #2777, opened 2022-04-07
provide some space between filter controls and selected terms.

stopfstedt:
fixes https://github.com/ilios/common/issues/2764


dartajax (approved):
that 1 rem is the only rem we need - looks great

PR #2779, opened 2022-04-07
overlay screen while course rolls over.

### LINKED ISSUE #2778: Indicator Needed During Course Rollover
I was testing a PR and decided to test course rollover functionality.

Steps ....
* pick a course with lots of stuff to rollover 
* click the rollover icon from the toolbar below the publication status changing drop-down
* change the name of the course by adding something like " New" to the course's name
* now that the course is being rolled over into the same year, leave the start date of the course as well
* click the "Done" button
* try to move around and mouse over and click other stuff during the creation of the new course

I did this, accidentally tried to cancel, selected a cohort (which I figured I had forgotten to do earlier), and then all of a sudden the screen changed to indicate my newly created course, the existence of which I had just tried to cancel.

Or rollover any course of your choosing in any manner. We need to indicate status of course rollover. Maybe make the form read-only until the action has been completed.

stopfstedt:
fixes #2778

### DIFF
--- a/addon/components/course-rollover.hbs
+++ b/addon/components/course-rollover.hbs
@@ -3,6 +3,11 @@
   {{did-insert (perform this.load) @course}}
   {{did-update (perform this.load) @course}}
 >
+  {{#if this.save.isRunning}}
+    <WaitSaving
+      @showProgress={{false}}
+    />
+  {{/if}}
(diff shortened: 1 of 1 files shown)


jrjohnson (commented):
I'd personally rather see the button spin instead of the overlay. Is that just me?

PR #2789, opened 2022-04-11
trim whitespace before filtering courses on title.

### LINKED ISSUE #2788: Adding a Blank Space &nbsp Should Not Allow A Course to be Rolled Over Into the Same Year
image

Now when I add a single blank space, it appears the course can now be rolled over into the same year.

This should not be the case and will cause problems with multiple courses having the same name causing confusion.

See below ...

image

stopfstedt:
fixes #2788


dartajax (approved):
nice trimming - prevents an issue we need to prevent - approved

Issue #2801, opened 2017-02-08
No Results Found Message Needed When Doing a Search in Sessions


dartajax:
I was doing a search for "Sch" to look for "Scheduled" Sessions and noticed that the error msg states that there are no sessions in this course, but of course, there are 6 sessions in this course.

It might be better to say "No results found" or something instead.  I'd be pretty scared to think my sessions had "disappeared".

![image](https://cloud.githubusercontent.com/assets/7493331/22717780/d7d53fdc-ed51-11e6-91e3-cbd3537763b4.png)

PR #2807, opened 2022-04-14
show "no search results" message in sessions list.

stopfstedt:
fixes https://github.com/ilios/common/issues/2801


dartajax (approved):
great stuff

PR #2819, opened 2022-04-21
Set a 9px minimum font size

jrjohnson:
a11y rules require 9px to be the smallest size, our original step was
calculating to 8px, on most displays where we have a 16px body font
size. Using max ensures that we keep a minimum of 9px while allowing it
to grow at larger base size values.


jrjohnson:
@dartajax this applies to the monthly calendar, is this new slightly larger size OK?

dartajax (approved):
seems fine to me - I re-sized the window a few times and clicked around - didn't notice anything awry

jrjohnson:
Thanks for testing, this is ready to merge.

PR #2820, opened 2022-04-21
Improve dashboard router transition

jrjohnson:
replaceWith changes to the new route without adding an entry in the
browser history, since this redirect happens without any interim content
it's the right API to use here.


dartajax (approved):
click tested - seems fine

PR #2832, opened 2022-04-27
Add pronouns to user model and component

jrjohnson:
Pronouns will be available in the API and should be reflected in both
the user model and the component we use to display detailed name
information.

Refs ilios/ilios#4020


stopfstedt (approved):
LGTM

PR #2837, opened 2022-05-02
MeSH search refactoring.

### LINKED ISSUE #1907: Search Results Should Clear When Search Box Is Cleared >> Course MeSH
To reproduce ...
* Search for "Brain"

image

* Review the results
* Click the check marker icon to clear the search criteria

image

The search results from the previous search remain. Can we have them clear when the criteria gets cleared?

stopfstedt:
- close search results and reset search value if user clicks outside of
search results
- enables live search, our backend can handle it.
- rm obsolete submit search button
- adds text label
- corrects styles of loading indicator and no search results found.

fixes #1907

### DIFF
--- a/addon-test-support/ilios-common/page-objects/components/mesh-manager.js
+++ b/addon-test-support/ilios-common/page-objects/components/mesh-manager.js
@@ -1,13 +1,17 @@
 import {
+  attribute,
   clickable,
   collection,
   create,
+  fillable,
   hasClass,
   isVisible,
   notHasClass,
   text,
+  triggerable,
+  value,
 } from 'ember-cli-page-object';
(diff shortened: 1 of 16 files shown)


dartajax (approved):
I like it.

jrjohnson (commented):
Looks great and is a wonderful improvement. My preference would be to wait until a few more characters were entered to start searching or else cache the results from the first letter and then filter down from there. We've done both of these elsewhere. Having it fire on the first entered letter and the fire again on each new letter so quickly is sending a lot of network requests when someone types slow. Maybe just debounce a little bit longer on changes to the query?

jrjohnson (approved):

Issue #2842, opened 2022-04-12
Another Space Saver - Bring Details Up Higher - Courses and Sessions Offerings 


dartajax:
Screen shots below ...

First - how it is ...
<img width="1270" alt="Screen Shot 2022-04-12 at 3 32 58 PM" src="https://user-images.githubusercontent.com/7493331/163065640-529f74eb-325a-48ac-9e4a-24a7cd528ada.png">

Second - how it might be better if we brought the details up higher ...

![image](https://user-images.githubusercontent.com/7493331/163066509-6bc72532-37bd-48d2-9a17-b10de704907c.png)

PR #2843, opened 2022-05-06
re-layouts offering list items grid.

stopfstedt:
brings details and timing info up to the same vertical level rather than
stacking
puts more emphasis onto start/end of multiday events and drops
"multiday" label since this seems kinda redundant.

image

image

fixes https://github.com/ilios/common/issues/2842


dartajax (approved):
I like this change.

jrjohnson (approved):

Issue #2844, opened 2020-01-28
Fixing Positioning of My Activities For the Next 60 Days Label


dartajax:
It should look more like My Materials (shown below).

![image](https://user-images.githubusercontent.com/7493331/73314430-ada66900-41e2-11ea-84a3-07f33cccb501.png)

My Materials has the report output (none in this case) lined up with the caption. My Activities is not lined up so well. See below. 

![image](https://user-images.githubusercontent.com/7493331/73314483-db8bad80-41e2-11ea-90a9-e53b30ffc452.png)

PR #2845, opened 2022-05-06
realigns styles of activities component with waag.

stopfstedt:
fixes https://github.com/ilios/common/issues/2844


dartajax (approved):
Between the time this started and now, we have already reconfigured how the My Materials looks; however this looks nice.

PR #2855, opened 2022-05-18
Consolidate more data loading

jrjohnson:
The faster API responses mean we can pre-load a bit more data on a
course in a reasonable chunk of time. This significantly improves the
experience navigating within a course.


dartajax (approved):
It's a bit hard to test completely in common but it seems fine to me.

jrjohnson:
Actually I'm going to close this, I have an improved version that is in the works.

PR #2872, opened 2022-05-26
Apply Consistent Style to Top Level Pages

jrjohnson:
I've added a mixin to apply a main border style to things that touch the
navigation and header so they'll be consistent and pleasing.


dartajax (dismissed):
suggestion only ...

image

PR #2882, opened 2022-06-07
New Color Palette 

jrjohnson:
New matching colors for Ilios. 

I've created a new set of base colors along with a few variations based on lightness increments to give us some range. Each color has a standard name and gets referred to by that name instead of a more generic variable. I've also consolidated some of our many colors into a few.

Ilios 8 4-2


dartajax (commented):
in common you don't see the side flyout menus and all that stuff - but it looks good to me - I was wondering if you could also make the horizontal bars on the calendar like the vertical ones but maybe slightly less bold - or both a little less bold and the same

dartajax (approved):
I approve this and will merge it.

PR #2918, opened 2022-06-30
Fix Course Loading Jump

jrjohnson:
Couple of changes to ensure that when a course is loaded there isn't a weird jump between different states.


stopfstedt (approved):
LGTM

PR #2923, opened 2022-07-01
Add a class to create sticky table headers

jrjohnson:
For reasons that are very web you can't apply the sticky to a thead
element, you have to apply it to the th. Since our colors were on the
thead and once they rolled down it disappeared I had to add the color to
the th and make some adjustments elsewhere. We don't really need this
anywhere in common, but possibly on a very long list of LMs so I added
the style there.


stopfstedt (approved):
click tested and code reviewed. LGTM.

Issue #2934, opened 2021-12-01
Extra Vertical Space On Last Instructor Badge


dartajax:
![image](https://user-images.githubusercontent.com/7493331/144302665-f3f64bf1-9f32-4acd-be85-3a77c30a794f.png)

PR #2935, opened 2022-07-06
remove bottom margin override for last item in list.

stopfstedt:
while this may add additional .5rem bottom margin to some lists,
it also fixes all styling issues with any tag-lists that sit on top
of the default list styles.

fixes https://github.com/ilios/common/issues/2934


jrjohnson:
I'm ok with this if @dartajax doesn't mind the extra space.

dartajax (approved):
unable to thoroughly test on common but looks good to me as far as logic and code is concerned

PR #2936, opened 2022-07-08
move prework materials into session material list on event details page

stopfstedt:


jrjohnson (commented):
Love it. The incidental change pulling the LM list items into a component is very nice and makes for much more readable tests.

PR #2954, opened 2022-07-18
chore: remove superfluous complexity in clear()

Trott:
`delete` does not work on return values from functions.
`Map.prototype.clear()` does not take an argument. (Perhaps
`Map.prototype.delete()` was intended instead?)

### DIFF
--- a/addon/services/loading-opacity-tracker.js
+++ b/addon/services/loading-opacity-tracker.js
@@ -12,7 +12,4 @@ export default class LoadingOpacityTrackerService extends Service {
   has(key) {
     return this.#opacities.has(key);
   }
-  clear() {
-    this.#opacities.clear();
-  }
 }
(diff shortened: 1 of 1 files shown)


addon/services/loading-opacity-tracker.js
```diff
@@ -12,7 +12,7 @@ export default class LoadingOpacityTrackerService extends Service {
   has(key) {
     return this.#opacities.has(key);
   }
-  clear(key) {
-    delete this.#opacities.clear(key);
+  clear() {
+    this.#opacities.clear();
   }
```
jrjohnson, line 17:
```suggestion
```
This isn't used anywhere and isn't tested, let's just remove it.

jrjohnson (changes requested):

jrjohnson:
Thanks!

PR #2956, opened 2022-07-22
Stack Single Event Material Icons

jrjohnson:
Pulls the required and type icon into a grid and stacks them together
making the entire set look like a bullet point and cleaning up the
visual clutter in this list making it easier to see what each LM is.


dartajax (dismissed):
How it is in the PR ...

image

I believe we need a bit of separation between former "Pre-work" stuff and current Session LM's.

Maybe something like this ...

image

dashed line between the former "Pre-work" and current session LM's and maybe a bit (just a bit) of vertical space to designate the difference between the two

jrjohnson:
I created a new issue https://github.com/ilios/common/issues/2959 to make this change @dartajax.

stopfstedt (approved):
LGTM

dartajax (approved):
I approve.

PR #2960, opened 2022-07-25
Use Resources for course instructor visualizations

stopfstedt:
round and round we go.

follow-up to #2955


stopfstedt:
merged into #2955

PR #2970, opened 2022-07-27
reworks course session-types/session-type visualizations

stopfstedt:
- replace any invocations of Resources outside of `@use` with async methods
- improves chart labeling by dropping meaningless percentage from the bars
- fleshes out test coverage


jrjohnson (approved):

dartajax (approved):
This is an improvement - will merge.

Issue #2971, opened 2022-07-27
Session Terms and Learning Materials Not Displayed Properly in Print Preview


dartajax:
Here is from my sample course ...

This is from Session One ...

<img width="1252" alt="Screen Shot 2022-07-27 at 1 16 57 PM" src="https://user-images.githubusercontent.com/7493331/181364319-05389983-6145-4857-8700-b39e77da5553.png">

There are two Terms and one Learning Material attached to this Session.

Here is the same Session in the Print Preview ...

![image](https://user-images.githubusercontent.com/7493331/181364731-ac96a49f-01a2-4f99-a634-daa788f311e1.png)

PR #2976, opened 2022-07-28
print view: adds session lms and vocab terms back to the page

stopfstedt:
fixes https://github.com/ilios/common/issues/2971


dartajax (approved):
I tested this and have determined that it looks to have been fixed.

jrjohnson (approved):

PR #2977, opened 2022-07-28
Allow users to track material status

jrjohnson:
Using the new API users can now mark the status of their learning
materials on the single event detail page.

WIP: 
- [x] Translations


dartajax:
I love it - can you possibly make the formerly known as Pre-work check boxes have a bit more heft?

image

stopfstedt (approved):
LGTM

jrjohnson:
I can't change the checkboxes @dartajax, they only get worse. Let's add this as a possible followup though.

PR #2984, opened 2022-08-01
Replace WaaG PreWork with materials in events

jrjohnson:
Instead of having a top Prework section in the week at a glance add the
prework titles and materials into each event which matches how we're
displaying this information on the single event view.

Fixes ilios/ilios#4229


stopfstedt (approved):
LGTM

saschaben:
remaining issues:
- links to LMs listed under (ex)pre-work modules are still active in WaaG - they should be inactive in order to require click thru to the full activity event listing with objectives etc.

PR #2987, opened 2022-08-01
reworks course visualizations vocabularies/terms

stopfstedt:
- use async methods instead of resources as applicable
- relabels bars and tooltips on vocabulary terms chart to show total minutes instead of the less meaningful percentage numbers
- improves test coverage


jrjohnson (approved):

dartajax (approved):
looks good to me

PR #2988, opened 2022-08-02
Add some space between events in week-at-a-glance

jrjohnson:
And remove the zebra stripes


dartajax (changes requested):
There is more vertical space between offerings that have LM's - look at the blue boxes for reference - zoomed in yes but I noticed this in normal view. 

image

dartajax (approved):

Issue #2997, opened 2022-08-09
Blank White Row in Session List


dartajax:
See screen shot for reference ...

![image](https://user-images.githubusercontent.com/7493331/183738783-107b8528-6a98-4870-aca2-752eea41b7bb.png)

PR #2998, opened 2022-08-09
Session grid header background color

stopfstedt:
fixes https://github.com/ilios/common/issues/2997

### DIFF
--- a/app/styles/ilios-common/components/sessions-grid-offering-table.scss
+++ b/app/styles/ilios-common/components/sessions-grid-offering-table.scss
@@ -8,7 +8,6 @@
   width: 95%;
 
   thead {
-    background-color: $blueMunsell;
     color: $white;
     position: sticky;
     top: 2rem;
--- a/app/styles/ilios-common/components/sessions-grid-offering-table.scss
+++ b/app/styles/ilios-common/components/sessions-grid-offering-table.scss
@@ -2,7 +2,7 @@
 
 .sessions-grid-offering-table {
   @include ilios-table-structure;
-  @include ilios-table-colors;
+  @include ilios-table-colors($blueMunsell);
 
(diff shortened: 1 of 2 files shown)


dartajax (approved):
looks good to me

app/styles/ilios-common/components/sessions-grid-offering-table.scss
```diff
@@ -2,7 +2,7 @@
 
 .sessions-grid-offering-table {
   @include ilios-table-structure;
-  @include ilios-table-colors;
+  @include ilios-table-colors($blueMunsell);
```
jrjohnson, line 5: I'm hesitant to accept this because it appears to duplicate what is on line 11 of this file in setting the background color in the thead of this table? I think maybe the change should be to apply the color to [the th element](https://github.com/ilios/common/pull/2998/files#diff-8886b7715030234c793a0d7caaca81cb9d697ace4bef827b820416d0d0f83ef7L16) and not supply this as an option to the mixin or to remove the declaration from [the head](https://github.com/ilios/common/pull/2998/files#diff-8886b7715030234c793a0d7caaca81cb9d697ace4bef827b820416d0d0f83ef7L11) and use the mixin.

jrjohnson (changes requested):

app/styles/ilios-common/components/sessions-grid-offering-table.scss
stopfstedt: i removed the redundant style declaration from row 11. that should fix this.

stopfstedt: :heavy_check_mark:

jrjohnson (approved):

PR #3006, opened 2022-08-10
improves input validation for learning material title.

### LINKED ISSUE #3005: Earlier Validation of Max Length in Course Learning Material Title is Needed
See the screen shot -- in this case I went through the rigamarole of entering all of the other values for my LM. I clicked "Done" before the validation message appears. It is my belief that the error message should be displayed as soon as the user reaches the maximum number of allowable characters.

This is consistent with other areas such as "Course ID", where the error appears as soon as the max is reached.

Screen Shot 2022-08-09 at 3 46 37 PM

stopfstedt:
fixes #3005

### DIFF
--- a/addon/components/learningmaterial-manager.hbs
+++ b/addon/components/learningmaterial-manager.hbs
@@ -17,7 +17,7 @@
             type="text"
             value={{this.title}}
             disabled={{this.save.isRunning}}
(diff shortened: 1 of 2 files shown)


dartajax (approved):
very helpful - I like this.

PR #3024, opened 2022-08-24
eliminate fieldset elements from single event details page.

### LINKED ISSUE #3023: Slight Improvement to Event Detail Appearance
The screen shots below don't quite get at what I'm trying to point out. The gist is move the details of the event detail over a bit to the right so they line up better with the stuff below - Materials and the other stuff below.

Here is how it is currently - logged in as student ...

image

I am suggesting something a bit more like ...

image

stopfstedt:
this results in consistent left-hand vertical alignment of all page
elements, at the expense of losing the left hand grey border.
seems worth it to me.

image

fixes #3023

### DIFF
--- a/addon/components/single-event.hbs
+++ b/addon/components/single-event.hbs
@@ -101,76 +101,72 @@
         <br>
       {{/if}}
     </div>
-    <fieldset>
-      <div class="single-event-learningmaterial-list" data-test-session-materials>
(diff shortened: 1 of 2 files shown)


dartajax (approved):
looks good to me

jrjohnson (approved):

Issue #3025, opened 2022-03-31
Need More Information Available to Select Course in Materials / All Materials


dartajax:
![image](https://user-images.githubusercontent.com/7493331/161158770-112d6522-db17-4b8f-abc3-94fc8ea2629f.png)

stopfstedt:
needs backend changes.

stopfstedt:
first impressions after prototyping this with actual data - cramming all this extra info into a simple drop-down is not working.

dashes as separators, it's horrid since course title, course year and/or external id all can contain dashes within.

![image](https://user-images.githubusercontent.com/1410427/186531202-65f41ec0-6a71-412b-9eb0-6d2095778891.png)

putting secondary info into parenthesis - a bit better, but still baffling.
![image](https://user-images.githubusercontent.com/1410427/186530459-477e53a0-33fb-4f82-b43e-6841ad4b6d2e.png)


this needs to be re-thought.

couple of ideas:

1. i think we could get away with putting _one_ of those two requested additional attribute into parenthesis after the course title, but not both.

2. we toss out the simple drop-down in favour of something more expressive, like the user search results or MeSH search results. will require a decent amount of work to craft a new UI component just for that, obviously.


in any case - back to the drawing board.

saschaben:
I think the right approach is some form of #3 above - identify/group/sort by academic year, and then provide course titles + parenthetical ext. ID as the clickable entities.

dartajax:
I would suggest maybe just the year in brackets would be good - maybe we can do that.

PR #3026, opened 2022-08-24
adds external id and year to courses dropdown.

stopfstedt:
fixes https://github.com/ilios/common/issues/3025

image


jrjohnson (approved):

jrjohnson:
Waiting for API change before merging.

dartajax (changes requested):
see screen shot below - for reference this is demo_student24 logged in ...

image

the controls look kind of squished when some of the concatenated course titles are longer - maybe we should put in a couple of &nbsp's in there - if not, I will approve this as is since I like the drop-down value concatenation in general especially since it will match the pattern used in other places - just a suggestion

Issue #1742, opened 2020-11-04
Remove agenda view from dashboard


stopfstedt:
per team discussion 11/03/2020.

nancirose:
Need additional information to address issue.
-Nanci, Michael, Stacey

PR #1743, opened 2020-11-04
removes agenda view from dashboard.

stopfstedt:
fixes #1742

PR #3030, opened 2022-08-26
rm agenda ("activities") view from dashboard.

stopfstedt:
fixes https://github.com/ilios/common/issues/1742
supersedes #1743


jrjohnson (approved):
I can't tell if this looks weird because of the button styles or because I'm used to there being 4 buttons, either way I approve and we'll have to think about styling or maybe now.

PR #3031, opened 2022-08-26
upgrade to ember-concurrency v2.3.2

stopfstedt:
this replaces task modifiers with async task API.


jrjohnson (approved):
As approved as it's gonna get I think.

PR #3044, opened 2022-08-30
adds icon to user search results to indicate disabled accounts.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4372

image

(will need a common release in order to take affect in the frontend UI that the linked ticket is referencing)


dartajax (approved):
can't really get to this in common - but from the screenshots, logic, and code, I will approve this - shared component seems fine to me - good to know when an account is disabled

PR #3057, opened 2022-08-31
Be more specific about peer deps

jrjohnson:
If we don't do this NPM messes with our dependency versions when we
update. Let's try this and see how it goes.


stopfstedt (approved):
i accept this.

PR #3058, opened 2022-08-31
gives visualization mixin a more generic name.

stopfstedt:
we're re-using those styles for non-course related visualizations, so
let's make sure this is correctly reflected in the name of the mixin.

refs https://github.com/ilios/frontend/pull/6842


stopfstedt:
@dependabot-bot rebase

jrjohnson (approved):

PR #3059, opened 2022-09-01
updates outdated import with current equivalent.

stopfstedt:
companion to https://github.com/ilios/frontend/pull/6845


dartajax (approved):
click tested using student token - seems fine to me

PR #3068, opened 2022-09-07
Reworks Learnergroup selection component

stopfstedt:
image

- makes it more evident what is selected (and what is not) by using checkboxes in the picker.
- use tag lists for selected groups that contain the full parentage of each group, making selected parent groups explicit.
- fixes state handling/loading bug that was preventing cascading selects on first click in the picker
- adds ctrl+click/shift-click support to allow selection/deselection of groups with children without cascading effect.

fixes https://github.com/ilios/ilios/issues/3905
fixes https://github.com/ilios/ilios/issues/3896


dartajax (commented):
I think I found a bug. See below ...

These are sub groups of BRIDGES Class of 2026 >> Aardvark Island
I am going to select Aardvark Island >> Not Enough Ants >> Never Enough Ants first.

Screen Shot 2022-09-07 at 2 24 37 PM

After performing the action shown above ...

image

I stated in my screen shot that it was correct - It actually is not correct - I apologize for that ....

I selected `Aardvark Island >> Not Enough Ants >> Never Enough Antz`
It appears that I selected `Not Enough Ants >> Aardvark Island >> Never Enough Antz` - the subs and parents are not in the correct order.

Now I return to select as shown below ...

Screen Shot 2022-09-07 at 2 28 08 PM

This should add the parent group so that now there are two groups selected but please review this screen shot for reference ...

image

The same group appears to have been added two times - This is pre-save though it should be pointed out. After Save ...

Screen Shot 2022-09-07 at 2 32 05 PM

Only one of the two groups (not three) were added to the offering. I wanted to add the child and the parent to the offering. Try this by directly selecting the parent to implicitly also select the child.

image

After pressing "Save" ...

Screen Shot 2022-09-07 at 2 36 30 PM


The child gets added but not the parent.

stopfstedt:
@dartajax i addressed the following bugs:

- corrected order of parent group titles  in selected groups tags (fixes sort order as well)
- de-duplicates selected groups
- prevent "knockout" effect that was yanking selected groups if they have selected descendants

please click-test again.

dartajax (commented):
All the things I reported in my previous comment have been fixed and look really nice ...

Here's another potential mini-caveat.

image

dartajax:
Now that we have check boxes, the default if a parent is selected is to have all children selected as well. That's fine but what if we just want the children and don't want to have to go back in a re-click each child. Each check box applies to that one group now so is that possible?

dartajax (approved):
I am good with this at least on my Shift using Mac.

dartajax (approved):

addon/components/learnergroup-tree.js
```diff
@@ -58,10 +54,13 @@ export default class LearnergroupTree extends Component {
   }
 
   @action
-  add(learnerGroup) {
-    if (this.selectable) {
-      this.args.add(learnerGroup);
-    }
+  add(learnerGroup, ev) {
+    this.args.add(learnerGroup, !(ev.ctrlKey || ev.shiftKey));
```
jrjohnson, line 58: Wow. That's 100x cleaner than I could have imagined. Nice!

addon/models/learner-group.js
```diff
@@ -336,7 +336,25 @@ export default class LearnerGroup extends Model {
       return [];
     }
     const allParents = await parent.getAllParents();
-    return [parent, ...allParents];
+    return [...allParents, parent];
+  }
+
+  async getAllParentTitles() {
```
jrjohnson, line 342: Why not `this.allParentTitles`?

jrjohnson (approved):

PR #3075, opened 2022-09-14
consider associated learner-groups in session publication check.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4287

### DIFF
--- a/addon-test-support/ilios-common/page-objects/components/session-publicationcheck.js
+++ b/addon-test-support/ilios-common/page-objects/components/session-publicationcheck.js
@@ -10,6 +10,7 @@ const definition = {
   offerings: text('[data-test-offerings]'),
   terms: text('[data-test-terms]'),
   objectives: text('[data-test-objectives]'),
+  learnerGroups: text('[data-test-learnergroups]'),
   unlink: {
     scope: '[data-test-unlink]',
   },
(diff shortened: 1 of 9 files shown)


dartajax (commented):
not quite sure about this one ... yet 

consider the following which allows publication to occur ... it says there are 4 things missing but there are actually 5 - this is an ILM so an "offering" is not required but because of the fact that it is an ILM, an "offering" is not technically required - in other because of the fact that this is an an ILM, the "offering" requirement has actually been fulfilled 

image

stopfstedt:
> not quite sure about this one ... yet
> 
> consider the following which allows publication to occur ... it says there are 4 things missing but there are actually 5 - this is an ILM so an "offering" is not required but because of the fact that it is an ILM, an "offering" is not technically required - in other because of the fact that this is an an ILM, the "offering" requirement has actually been fulfilled
> 
> image

is this a regression or a new bug? if it's an existing issue then please file a new ticket. thanks.

stopfstedt:
@dartajax to answer my own question - this has nothing to do with the issue at hand, and is also an existing problem if you want to call it as such


ILM publication check:
image


please file a new ticket on this and describe what the issue is and how it should be addressed.

stopfstedt:
i'm withdrawing this PR. turns out the idea is half-baked. while this works well for ILMs, the information provided for sessions with offerings is incomplete at best, and misleading at worst. 

there's no benefit to tell saying "there are 5 learner groups" in the publication check if all of them are attached to the same offering, and none are attached to 3 other offerings.

PR #3079, opened 2022-09-19
Remove array prototype extensions

jrjohnson:
Ember data has deprecated and ember will soon deprecate extensions to the Array prototype. By adding the `ember-disable-prototype-extensions` package we can ensure that these extensions are not used anywhere in our code and I've gone through and replaced every usage.

For convenience I added an array helpers utility set of functions that does some of the things we used to get from ember, in other cases a better alternative exists already in JS.

### DIFF
--- a/addon/utils/array-helpers.js
+++ b/addon/utils/array-helpers.js
@@ -1,0 +1,118 @@
+import { get } from '@ember/object';
+import { DateTime } from 'luxon';
+
+export function mapBy(arr, path) {
+  if (arr !== null && typeof arr === 'object') {
+    if ('slice' in arr) {
+      arr = arr.slice();
+    }
+  }
+  if (!Array.isArray(arr)) {
+    return arr;
+  }
+  return arr.map((item) => {
+    if (item === undefined) {
+      return undefined;
+    }
+    if (item === null) {
+      return null;
+    }
+
+    return get(item, path);
+  });
+}
+
+export function sortByString(arr, path) {
+  if (arr !== null && typeof arr === 'object') {
+    if ('slice' in arr) {
+      arr = arr.slice();
+    }
+  }
+
+  if (!Array.isArray(arr)) {
+    return arr;
+  }
+
+  return arr.sort((objA, objB) => {
+    const a = get(objA ?? {}, path);
+    const b = get(objB ?? {}, path);
+
+    if (a === b) {
+      return 0;
+    }
+
+    if (a === undefined || a === null) {
+      return 1;
+    }
+
+    if (b === undefined || b === null) {
+      return -1;
+    }
+
+    return a.localeCompare(b);
+  });
+}
+
+export function sortByDate(arr, path) {
+  if (arr !== null && typeof arr === 'object') {
+    if ('slice' in arr) {
+      arr = arr.slice();
+    }
+  }
+
+  if (!Array.isArray(arr)) {
+    return arr;
+  }
+
+  return arr.sort((objA, objB) => {
+    const a = get(objA ?? {}, path);
+    const b = get(objB ?? {}, path);
+
+    if (a === b) {
+      return 0;
+    }
+
+    if (a === undefined || a === null) {
+      return 1;
+    }
+
+    if (b === undefined || b === null) {
+      return -1;
+    }
+    const d1 = DateTime.fromJSDate(a);
+    const d2 = DateTime.fromJSDate(b);
+
+    return d1.toMillis() - d2.toMillis();
+  });
+}
+
+export function sortByNumber(arr, path) {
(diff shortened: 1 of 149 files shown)


addon/utils/array-helpers.js
```diff
@@ -167,3 +167,29 @@ export function uniqueValues(arr) {
...
 }
+
+export function findBy(arr, key, searchValue) {
+  if (arr !== null && typeof arr === 'object') {
+    if ('slice' in arr) {
+      arr = arr.slice();
+    }
+  }
+
+  if (!Array.isArray(arr)) {
```
stopfstedt, line 116: this does not track for me at all.
if the given input is not an array, or cannot be extracted into an array, then _it itself_ is returned. 
that seems like a bug to me.
consider this test passing

```js
  test('findBy when input is not array and cannot be converted into an array', function (assert) {
    const arr = {};
    const result = findBy(arr, 'id', 3); // doesn't matter what filters you apply here, you always get the input "array" back.
    assert.strictEqual(result, arr);
  });
```

i think an Error should be raised here instead.

addon/utils/array-helpers.js
```diff
@@ -193,3 +193,30 @@ export function findBy(arr, key, searchValue) {
...
 }
+
+export function filterBy(arr, key, searchValue) {
+  if (arr !== null && typeof arr === 'object') {
+    if ('slice' in arr) {
+      arr = arr.slice();
+    }
+  }
+
+  if (!Array.isArray(arr)) {
```
stopfstedt, line 142: should raise an Error or Exception instead.

see comment about `findBy()`.

addon/utils/array-helpers.js
```diff
@@ -0,0 +1,252 @@
...
+}
+
+export function uniqueBy(arr, key) {
+  if (arr !== null && typeof arr === 'object') {
+    if ('slice' in arr) {
+      arr = arr.slice();
+    }
+  }
+
+  if (!Array.isArray(arr)) {
```
stopfstedt, line 65: should throw Error/Exception, see comment about `findBy()`.

addon/utils/array-helpers.js
```diff
@@ -0,0 +1,252 @@
+import { get } from '@ember/object';
+import { assert } from '@ember/debug';
+
+export function mapBy(arr, path) {
+  if (arr !== null && typeof arr === 'object') {
+    if ('slice' in arr) {
+      arr = arr.slice();
+    }
+  }
+  if (!Array.isArray(arr)) {
```
stopfstedt, line 10: should throw Error/Exception, see comment about `findBy()`.

addon/utils/array-helpers.js
```diff
@@ -0,0 +1,252 @@
...
+
+//stolen from https://github.com/emberjs/ember.js/blob/464e694afd611e2203759e5f76a14c7bfb023006/packages/%40ember/-internals/runtime/lib/mixins/array.ts#L1371
+export function sortBy(arr, sortKeys) {
+  if (arr !== null && typeof arr === 'object') {
+    if ('slice' in arr) {
+      arr = arr.slice();
+    }
+  }
+
+  if (!Array.isArray(arr)) {
```
stopfstedt, line 33: should throw Error/Exception, see comment about `findBy()`.

addon/utils/array-helpers.js
```diff
@@ -0,0 +1,252 @@
...
+}
+
+export function uniqueValues(arr) {
+  if (arr !== null && typeof arr === 'object') {
+    if ('slice' in arr) {
+      arr = arr.slice();
+    }
+  }
+
+  if (!Array.isArray(arr)) {
```
stopfstedt, line 102: should throw Error/Exception, see comment about `findBy()`.

addon/utils/array-helpers.js
```diff
@@ -116,3 +116,54 @@ export function sortByNumber(arr, path) {
...
+
+  const seen = new Set();
+  const rhett = [];
+  arr.forEach((item) => {
+    if (item === undefined) {
+      rhett.push(undefined);
+      return;
+    }
+    if (item === null) {
+      rhett.push(null);
```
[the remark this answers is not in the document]
jrjohnson: Awesome catch, will fix.

addon/utils/array-helpers.js
stopfstedt: whatever we end up doing - return a blank array in case of bad input, raise an error, or whatever else - the function should not return itself if it's not an array/array-like.

perhaps returning an empty array and adding a type-checking [assertion](https://developer.mozilla.org/en-US/docs/Web/API/console/assert) here might be the least invasive solution.

stopfstedt (approved):

PR #3080, opened 2022-09-19
Course LMs style improvements

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4399
refs https://github.com/ilios/ilios/issues/4400


dartajax (approved):
Ein besserung.

PR #3087, opened 2022-09-26
replace ObjectProxy with component

stopfstedt:
this is hokey, let's replace it.

while at it, convert the `has-route` helper to use class syntax.


jrjohnson (approved):

stopfstedt:
> not sure - I was going to just merge this without testing but I noticed uploading a learning material file fails - is that because this is a common PR? I just tested a different PR yesterday and this seemed to work - also you can `Session Type - (inactive)` flash really quick upon refresh - I hadn't seen that before either although it's probably unrelated

rebased against `master`, this should work now.

dartajax (approved):
works fine now - gonna merge

PR #3089, opened 2022-09-28
converts collections to arrays before accessing them as such.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4412


dartajax (approved):
I am able to upload and modify LM's, URL's and citations where appropriate. So ... looks good to me.

dartajax:
ah missed my chance - all good

PR #3091, opened 2022-09-29
Replace moment with luxon

jrjohnson:
While I started out restricting this to tests the fact that the first day of the week was handled by the localization layer in moment quickly became a problem and I was forced to make some application code changes as well so calendars will need to be tested. I did this one test at a time for the most part for easier review, but looking at all the commits maybe I should squash them?


dartajax (commented):
not sure about this one - when I try to pull up all events, no events show up - there are school events that should show up - I am using the token of `demo_dev1`.

stopfstedt (changes requested):
the week view in the calendar is not rendering events correctly. latest FF/Linux.

jrjohnson:
Looks like the same issue caused both problems, fixed now, ready for another review.

stopfstedt (approved):
LGTM

dartajax (approved):
looks good to me

PR #3099, opened 2022-10-07
Expand dashboard waag and materials to fill space

jrjohnson:
Now that we've dropped the reports and courses blocks from the dashboard we should use the entire space.


dartajax (approved):
I like it.

PR #3102, opened 2022-10-10
fixes regression in sorting of offerings in list.

stopfstedt:
the first element of an EmberData collection cannot be accessed via array index, b/c it's not an array (yet).
reverting the change introduced in b7b791.

fixes https://github.com/ilios/ilios/issues/4437

### DIFF
--- a/addon/utils/offering-date-block.js
+++ b/addon/utils/offering-date-block.js
@@ -57,7 +57,7 @@ const OfferingDateBlock = OfferingBlock.extend({
 const OfferingTimeBlock = OfferingBlock.extend({
   init() {
     this._super(...arguments);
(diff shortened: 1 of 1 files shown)


stopfstedt:
complete lack of test coverage on this component. i'm writing these up today, but feel free to merge if this passes muster for the sake of getting this fix out of the door.

jrjohnson (approved):
Accepted as a nice quick fix. We probably need to completely re-write sort here, will wait for the new tests. I'd recommend focusing on acceptance tests here because I think we may need to completely re-write the underlying backing object + component code but output the same HTML.

dartajax (approved):
the sorting issue seems to be fixed up - works for me

PR #3119, opened 2022-10-18
Remove disable prototype extensions addon

jrjohnson:
We've removed these extensions, I'm removing this safety check for now in order to allow ember data to be updated. When prototype extensions is enabled Ember.A() becomes a noop, but with this addon installed it clobbers the RecordArray objects from ember-data. Wile we don't use Ember.A ourselves it is used in ember-composable-helpers. This is a temporary solution until Ember is updated to prevent it from clobbering ED arrays, or ember-composable-helpers is updated to not use Ember.A(). PRs are in for both of these changes, but getting them merged and released is going to take some time.


package.json
```diff
@@ -39,13 +39,13 @@
...
     "@formatjs/intl-relativetimeformat": "^11.0.3",
     "@fortawesome/ember-fontawesome": "^0.4.1",
-    "@fortawesome/free-brands-svg-icons": "^6.2.0",
-    "@fortawesome/free-regular-svg-icons": "^6.2.0",
-    "@fortawesome/free-solid-svg-icons": "^6.2.0",
-    "@fortawesome/pro-duotone-svg-icons": "^6.2.0",
-    "@fortawesome/pro-light-svg-icons": "^6.2.0",
-    "@fortawesome/pro-regular-svg-icons": "^6.2.0",
-    "@fortawesome/pro-solid-svg-icons": "^6.2.0",
+    "@fortawesome/free-brands-svg-icons": "6.2.0",
```
stopfstedt, line 42: i assume pinning those versions is intentional.

stopfstedt (approved):

package.json
jrjohnson: Nope. I have no idea why that happened. Ugh. I'll look into it.

dartajax:
sorry if I merged that too early @jrjohnson

PR #3123, opened 2022-10-24
Use our internationalization library to format dates

jrjohnson:
This helps with cross browser support for these tests.


addon/components/date-picker.js
```diff
@@ -61,4 +62,9 @@ export default class DatePickerComponent extends Component {
       this._flatPickerInstance.destroy();
     }
   }
+
+  async onChange(date) {
+    await this.args.onChange(date);
+    await next(() => {});
```
stopfstedt, line 68: :crossed_fingers:

stopfstedt (approved):

PR #3133, opened 2022-11-03
Updating Timed Release LM Message

dartajax:
shortening the timed release message for easier reading

### DIFF
--- a/translations/en-us.yaml
+++ b/translations/en-us.yaml
@@ -338,7 +338,7 @@ general:
   timedReleaseNoSchedule: "Available immediately when published"
   timedReleaseOnlyEndDate: "Available until {endDate}"
   timedReleaseOnlyStartDate: "Available: {startDate}"
-  timedReleaseStartAndEndDate: "Available: {startDate} and available until {endDate}"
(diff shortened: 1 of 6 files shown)


translations/en-us.yaml
```diff
@@ -338,7 +338,7 @@ general:
   timedReleaseNoSchedule: "Available immediately when published"
   timedReleaseOnlyEndDate: "Available until {endDate}"
   timedReleaseOnlyStartDate: "Available: {startDate}"
-  timedReleaseStartAndEndDate: "Available: {startDate} and available until {endDate}"
```
jrjohnson, line 341: Does this need to change in the French and Spanish translations as well?

jrjohnson (changes requested):
You will also need to update the language in the `single-event-learningmaterial-list-item` test, let me know if you want to pair on that change.

translations/en-us.yaml
dartajax: I had tested this in Spanish before starting on this work and found it was fine - same with French. This issue was only in English.

jrjohnson (approved):

PR #3144, opened 2022-11-13
Increase timeout for floating tests

jrjohnson:
These are taking a very long time to run, not sure why but we can kick up the timeout so they're allowed to finish.


jrjohnson:
Nope, this isn't the issue. The problem may be ember-data. Investigating.

PR #3145, opened 2022-11-13
Pin ember-data at 4.7.3

jrjohnson:
Latest ember-data does not get along with ember-concurrency. Working on this upstream, but for now let's pin it at this working version to ensure we can keep working.


jrjohnson:
Going to merge this myself so everything else can get re-based and pass.

PR #3147, opened 2022-11-15
rewires/corrects "scroll to publicationcheck" behaviour.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4481


jrjohnson (changes requested):
We actually [have a modifier](https://github.com/ilios/common/blob/master/addon/modifiers/scroll-into-view.js) that handles this without needed component JS `{{scroll-into-view}}` can be added anywhere and it will do this same thing.

jrjohnson (approved):

PR #3149, opened 2022-11-16
Fix token refresh and expiration

jrjohnson:
Couple of issues fixed here, first we weren't respecting our own configuration, second we weren't scheduling token expiration. Both of these were causing repeated error messages when someone attempted to do some work with an expired token which shows up as authentication errors in our logs.


stopfstedt (approved):
LGTM

PR #3150, opened 2022-11-16
Improve Login Experience

jrjohnson:
I've aligned our configuration with the frontend to remove token refresh and updated the login itself to use the token we pass and not a refreshed version. We still attempt to use the token to get a new token to ensure the passed token is valid and when there are errors in that process they will now display correctly.

### DIFF
--- a/tests/dummy/app/controllers/login.js
+++ b/tests/dummy/app/controllers/login.js
@@ -21,9 +21,10 @@ export default class LoginController extends Controller {
         },
       });
       if (response.ok) {
-        const obj = await response.json();
         const authenticator = 'authenticator:ilios-jwt';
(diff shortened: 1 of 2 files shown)


tests/dummy/app/controllers/login.js
```diff
@@ -21,9 +21,10 @@ export default class LoginController extends Controller {
         },
       });
       if (response.ok) {
-        const obj = await response.json();
         const authenticator = 'authenticator:ilios-jwt';
-        this.session.authenticate(authenticator, { jwt: obj.jwt });
+        this.session.authenticate(authenticator, { jwt: this.jwt });
+      } else {
+        this.error = await response.text();
```
stopfstedt, line 27: :+1:

stopfstedt (approved):

PR #3165, opened 2022-11-22
Removes dependency on ember-composable-helpers

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4492
refs https://github.com/ilios/ilios/issues/4491


addon/helpers/sort-by.js
```diff
@@ -0,0 +1,192 @@
+// taken from Ember Composable Helpers (https://github.com/DockYard/ember-composable-helpers), then modified.
```
jrjohnson, line 1: Wondering if we can import the sorter from https://github.com/ilios/common/blob/master/addon/utils/array-helpers.js#L28 and use that here? Maybe another step.

addon/helpers/queue.js
```diff
@@ -0,0 +1,23 @@
+// taken from Ember Composable Helpers (https://github.com/DockYard/ember-composable-helpers), then modified.
+import { helper } from '@ember/component/helper';
+import isPromise from '../utils/is-promise';
+
+export default helper(function queue(actions = []) {
+  return function (...args) {
+    let invokeWithArgs = function (acc, curr) {
+      if (isPromise(acc)) {
```
jrjohnson, line 8: I hate that this is here, random promise gets resolved in weird way. Just complaining, no need to change it in case we're using it somehwere.

jrjohnson (approved):
I didn't look at the test files, but looked at app code and it LGTM.

addon/helpers/sort-by.js
stopfstedt: good idea, but should be a follow-up item. will create a corresponding ticket.

addon/helpers/queue.js
stopfstedt: checked, and we're not feeding it any Promises anywhere. removed that feature from the  helper :heavy_check_mark:

PR #3237, opened 2023-02-14
resets filters to "Course/Session Type" when filters are hidden.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4654

### DIFF
--- a/addon/controllers/dashboard/calendar.js
+++ b/addon/controllers/dashboard/calendar.js
@@ -82,7 +82,7 @@ export default class DashboardCalendarController extends Controller {
     if (this.showFilters) {
       this.showFilters = false;
       this.academicYear = null;
-      this.courseFilters = null;
+      this.courseFilters = true;
     } else {
       this.showFilters = true;
     }


jrjohnson (approved):

dartajax (changes requested):
I hate to be that guy but yes upon first opening up and viewing the filters, the default should be `"Course / Session Type"` but if the user has selected the other filter `"Program Detail"` and then clicks `"Hide Filter"` and then clicks `"Show Filters"`, their selection of `"Program Detail"` should still apply. In other words, keep the user's selection if they are turning filters on and off. Doable?

stopfstedt:
> I hate to be that guy but yes upon first opening up and viewing the filters, the default should be `"Course / Session Type"` but if the user has selected the other filter `"Program Detail"` and then clicks `"Hide Filter"` and then clicks `"Show Filters"`, their selection of `"Program Detail"` should still apply. In other words, keep the user's selection if they are turning filters on and off. Doable?

doable but disagree. filters should reset to their initial state when you hide them. flag as team discussion if you want.

PR #3238, opened 2023-02-15
reverses sort order of academic years in courses filter.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4621


dartajax (approved):
looks good to me

PR #3239, opened 2023-02-16
[breaking] Refactor Styles

jrjohnson:
Sass [modules](https://sass-lang.com/documentation/at-rules/use) are the preferred way to link our mixins and variables together in new versions of sass. This is the least amount of change I could make work while still making an attempt to align with this new paradigm.

This will require a similar total restructuring of styles on frontend so this is a breaking change.


jrjohnson:
This is now ready for review and possible merging.

stopfstedt (approved):
spot-checked code changes and click-tested UI. LGTM.

PR #3240, opened 2023-02-16
adds option for showing an edit-icon edit-in-place component

stopfstedt:
image

we'll need this in at least one place (namely, the subject report header) to visually indicate edit-in-place fields better - to set them apart from "regular" links.

the span to button conversion is unrelated to this (aka "while i was in there").


jrjohnson (approved):
LGTM

PR #3247, opened 2023-02-27
Replace access to string with toString in tests

jrjohnson:
These are the results of htmlSafe(ing) a string and the API for accessing the string itself has changed in the latest versions of ember.


jrjohnson:
Test failures are unrelated, this can be merged with failures which will be fixed with updated dependencies elsewhere.

stopfstedt (approved):

PR #3252, opened 2023-03-01
Small tweaks to course rollover date picker

jrjohnson:
Using luxon here to compare dates in the hopes of clearing up some issues in Safari and also following a recommendation from my IDE to not attempt to optional chain on a private property and instead check the existence the old fashioned way.


stopfstedt (approved):
LGTM

PR #3253, opened 2023-03-01
Use luxon from JSON:API date conversion utility

jrjohnson:
Removing moment one bit at a time, but also added some docs because I hate this little utility and it makes me cry every time I see it.


stopfstedt (approved):
LGTM. the real win here are the code comments.

PR #3267, opened 2023-03-07
Left Align Link Buttons

jrjohnson:
Just like links, link-buttons should default to a left alignment. I removed any duplicate left-align styles and added center alignment where it seemed warranted. I also removed a double style assignment for the copy button on offering url.


stopfstedt (approved):
click-tested and reviewed, LGTM

PR #3269, opened 2023-03-10
replace usage of await helper from templates with Resource-based async data loading process.

stopfstedt:
refs https://github.com/ilios/ilios/issues/4651

### DIFF
--- a/tests/acceptance/dashboard/calendar-test.js
+++ b/tests/acceptance/dashboard/calendar-test.js
@@ -777,7 +777,7 @@ module('Acceptance | Dashboard Calendar', function (hooks) {
   });
 
   const pickTerm = async function (i) {
-    return await click(find(`.vocabularyfilter li:nth-of-type(${i}) [data-test-target]`));
+    return await click(find(`.vocabularyfilter [data-test-filter-checkbox-target-id="${i}"]`));
   };
 
   test('test term filter', async function (assert) {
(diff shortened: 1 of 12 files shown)


addon/components/dashboard/calendar.hbs
```diff
@@ -87,6 +87,7 @@
                           @checked={{includes type.id @selectedSessionTypeIds}}
                           @add={{fn @addSessionTypeId type.id}}
                           @remove={{fn @removeSessionTypeId type.id}}
+                          @id={{type.id}}
```
jrjohnson, line 90: I'm a little worried that this being called `id` is going to make a future problem where I decide every checkbox needs an ID and use this passed property for it, however it's not unique. Maybe do `@id="seession-type-{{type.id}}"` (and the other places this gets set) or you could set `data-test-filter-checkbox-id={{type.id}}` directly here and ensure `...attributes` gets called on the `<Dashboard::FilterCheckbox` component.

jrjohnson (changes requested):

addon/components/dashboard/calendar.hbs
stopfstedt: i renamed the argument to `@targetId` and the `data-test-` attribute to `data-test-filter-checkbox-target-id`. i hope this passes muster now.

tests/acceptance/dashboard/calendar-test.js
```diff
@@ -807,9 +807,9 @@ module('Acceptance | Dashboard Calendar', function (hooks) {
     await page.visit({ show: 'calendar', view: 'week' });
     await showFilters();
     assert.strictEqual(page.calendar.weeklyCalendar.events.length, 2);
+    //await pauseTest();
```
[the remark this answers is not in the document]
stopfstedt: removed. :heavy_check_mark:

jrjohnson (approved):

PR #3273, opened 2023-03-13
Be Explicit About Days With No Events

jrjohnson:
For users of assistive technology it can be confusing to view our calendar which contains a list of days, but may not contain any events. They may no know if there is truly nothing on the calendar or if we've coded our site poorly. This a11y fix ensures that days without events contain a "no events" marker along with being visually empty.


stopfstedt (approved):
LGTM

PR #3276, opened 2023-03-15
Extract Calendar Filters into Component

jrjohnson:
Putting all of these in a component while removing the loader pattern. Should make it easier to make more changes in the future.


stopfstedt (approved):
LGTM

PR #3288, opened 2023-03-22
rm await from print course template.

stopfstedt:
replace it with Resources and use a sub-component for sessions.


jrjohnson (approved):
🔥, I thought this might be way more destructive. Nice.

dartajax:
Canary in a coal mine

PR #3291, opened 2023-03-23
replace enable-pull-request-automerge GH action with GH CLI equivalent.

stopfstedt:
according the action's author, it is no longer necessary to use it. 

see https://github.com/peter-evans/enable-pull-request-automerge#usage

refs https://github.com/ilios/common/pull/3290


.github/workflows/update-transitive-dependenies.yaml
```diff
@@ -35,8 +35,7 @@ jobs:
...
     - name: Enable Pull Request Automerge
       if: steps.cpr.outputs.pull-request-operation == 'created'
-      uses: peter-evans/enable-pull-request-automerge@v3
-      with:
-        token: ${{ secrets.ZORGBORT_TOKEN }}
-        pull-request-number: ${{ steps.cpr.outputs.pull-request-number }}
-        merge-method: merge
+      run: gh pr merge --merge --auto "$pr_number"
+      env:
+        GH_TOKEN: ${{ secrets.ZORGBORT_TOKEN }}
```
jrjohnson, line 40: Let's see if we can do this without a token now that it's not an action.

.github/workflows/update-transitive-dependenies.yaml
```diff
@@ -35,8 +35,7 @@ jobs:
...
+        pr_number: ${{ steps.cpr.outputs.pull-request-number }}
```
jrjohnson, line 41: I'm not certain about this plus the $pr_number in the `run` part. I don't think we need both and looking at the github docs we might not need either since this action targets the branch of the workflow.

jrjohnson (changes requested):

dartajax:
Except for the `Edge` case, it looks like this is ready to roll with resolved change requests and healthy conversation and analysis.

PR #3303, opened 2023-03-27
replace relative import paths with absolute paths.

stopfstedt:
we've been using both forms. let's standardize around this one.


jrjohnson:
I'm on the fence here. I'm worried the absolute path form only works because of some ember magic which may go away with embroider or we may need to configure NPM to understand where these may come from in order to get auto completion and such. Let's chat about it, I'll be easy to convince, I'm just not ready to pull the trigger without some discussion.

stopfstedt:
per our convo - let's move forward with this.

PR #3341, opened 2023-05-01
replace computed properties

stopfstedt:


jrjohnson (approved):
This is fantastic. I wondered if we should make the learner groups somehow part of the data so the sort components don't need async, but I think this is the right way to go.

PR #3406, opened 2023-06-15
Replace ResolveAsyncValue in models

jrjohnson:
The TrackedAsyncData provides a better way to manage asynchronous state, however it is more verbose and in order to be auto tracked it requires each step in an async getter chain to be explicitly coded. I focused on the testable output of the non-private getters for this change and I think I've replaced our original resources with these new getters. Long term we probably need to remove many of the public getters on models because doing things this way where we can't know exactly where in an async chain a change may happen makes for some verbose and difficult to read code.

### DIFF
--- a/addon/models/learner-group.js
+++ b/addon/models/learner-group.js
@@ -164,7 +164,7 @@ export default class LearnerGroup extends Model {
   }
 
   @cached
-  get _grandChildrenData() {
+  get _allDescendantsData() {
     if (!this._childrenData.isResolved) {
       return null;
     }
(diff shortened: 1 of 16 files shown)


addon/models/learner-group.js
```diff
@@ -192,44 +265,83 @@ export default class LearnerGroup extends Model {
...
     }
-    if (!this._parent?.allParents) {
-      return undefined;
+    const grandparents = await parent.getAllParents();
+
+    return [parent, ...grandparents];
+  }
+
+  @cached
+  get _grandparentsData() {
```
stopfstedt, line 314: suggesting name change to `_allAncestorsData()`.

stopfstedt (changes requested):
Two minor suggestions for change, otherwise this looks fine to me (at least from what i can see with my bleeding eyes).

addon/models/learner-group.js
```diff
@@ -107,15 +163,26 @@ export default class LearnerGroup extends Model {
     return ++offset;
   }
 
+  @cached
+  get _grandChildrenData() {
```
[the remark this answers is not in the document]
jrjohnson: ✅

addon/models/learner-group.js
jrjohnson: ✅

stopfstedt (approved):

PR #3415, opened 2023-06-27
Replace autodetected mirage models

jrjohnson:
The auto detector for mirage will not work anymore in ember-data v5. Instead we can just manually build our models and keep them up to date.


jrjohnson:
I'm seeing an issue with one of our tests in Percy. Investigating.

jrjohnson:
Thanks percy, that was a real bug! I tested against frontend as well, all seems in order. This is ready!

PR #3422, opened 2023-06-28
Limit or remove calls to toArray

jrjohnson:
This method is still present on ember-data relationships, but is deprecated and we shouldn't call it.


stopfstedt (approved):

jrjohnson:
embroider optimized failures reported upstream, merging anyway.

PR #3429, opened 2023-07-03
Clear Ember Data Deprecations

jrjohnson:
Broken down into small commits, this clears all the remaining deprecations before we can upgrade to v5 of ember data. 🎉 It's a big changeset, but most of these changes are in tests and therefor, I hope, self testing.


stopfstedt (approved):
outstanding. :fire:

jrjohnson:
Self merging because the failing embroider test has been reported upstream.

PR #3449, opened 2023-07-12
Calendar loading state indicator a11y improvements

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4773


jrjohnson (approved):
This is hugely more accessible, I'm not sure how I feel about the title being used for this, but it seems to work very well and sometimes I like it a lot and sometimes it's almost too fast and a bit jarring. I'm good to go with this and maybe adjust later, but would appreciate any input from @dartajax

dartajax (approved):
I am fine with this. It does look a bit weird since it is different; but aside from that, the message appearing briefly where the date range will be once loaded seems fine to me - a bit cleaner even.

PR #3450, opened 2023-07-12
prevent overlap by giving sticky header a higher z-index.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4119


dartajax (approved):
this is great

PR #3455, opened 2023-07-13
a11y - use <nav> element for pagination.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4765


jrjohnson (commented):
Good change, but something will have to be done to make this unique I think. a11y tests are complaining.

stopfstedt:
> Good change, but something will have to be done to make this unique I think. a11y tests are complaining.

corrected by providing separate a11y-labels for top and bottom nav.

jrjohnson (approved):

PR #3457, opened 2023-07-14
rm back-link from event details page.

stopfstedt:
this came up as unnecessary and potentially confusing during a a11y audit. begone!

fixes https://github.com/ilios/ilios/issues/4774

### DIFF
--- a/addon/controllers/events.js
+++ b/addon/controllers/events.js
@@ -1,9 +1,8 @@
 import Controller from '@ember/controller';
-import { getOwner } from '@ember/application';
+import { getConfig } from '@embroider/macros';
 
 export default class EventsController extends Controller {
   get showBackLink() {
-    const appConfig = getOwner(this).resolveRegistration('config:environment');
-    return !!appConfig.showHistoryBackLink;
+    return !!getConfig('showHistoryBackLink');
   }
 }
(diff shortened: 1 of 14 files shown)


stopfstedt:
i'm slapping a `do not merge` label on this since the companion PR https://github.com/ilios/lti-dashboard/pull/1373 is currenly failing.

jrjohnson (approved):

dartajax:
`DO NOT MERGE` label attached to this ticket - will merge once removed

addon/controllers/events.js
```diff
@@ -0,0 +1,9 @@
+import Controller from '@ember/controller';
+import { getOwner } from '@ember/application';
+
+export default class EventsController extends Controller {
+  get showBackLink() {
+    const appConfig = getOwner(this).resolveRegistration('config:environment');
```
jrjohnson, line 6: I *think* the way to do this has changed to [getConfig](https://github.com/embroider-build/embroider/tree/main/packages/macros#getownconfig-getconfig-and-getglobalconfig), but I haven't messed with it a lot. I'm ok with this solution, it works for sure, but we're already importing and using these macros other places so it's worth investigating.

jrjohnson (changes requested):

jrjohnson (approved):

stopfstedt:
i'm starting over on this.

PR #3460, opened 2023-07-20
rm scrolling feature from calendar views.

stopfstedt:
in the current implementation, the top navigation is pushed out of view when the daily or weekly calendar is loaded on the dashboard. this violates the first principle of Dave-Driven Design (TM) - "Don't confuse the user".
Since we don't have the level of control over the scrolling behaviour that would allow us to correct this issue, or at least i can't figure out how to do this, our only other option is to remove scroll-into-view functionality from the calendars altogether.

fixes https://github.com/ilios/ilios/issues/4867

### DIFF
--- a/addon/components/daily-calendar.hbs
+++ b/addon/components/daily-calendar.hbs
@@ -20,8 +20,6 @@
   <div
     class="day"
     tabindex="0"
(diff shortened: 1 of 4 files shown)


jrjohnson (approved):
‼️ 🚫

PR #3461, opened 2023-07-20
style the event title like an h3, even tho it's a h2.

stopfstedt:
fixes https://github.com/ilios/ilios/issues/4871

### DIFF
--- a/app/styles/ilios-common/components/single-event.scss
+++ b/app/styles/ilios-common/components/single-event.scss
@@ -25,7 +25,7 @@
     grid-column: 1 / -1;
 
     h2 {
-      @include m.ilios-heading-h1;
+      @include m.ilios-heading-h3;
       font-weight: normal;
 
       em {


dartajax (approved):
It's better indeed - was the h2 actually being passed as an h1 previously and now is being passed as h3? I imagine my question doesn't make sense but I am looking at the code to the right. Just curious ...

image

Anyway, @dartajax approves - great change.

stopfstedt:
> It's better indeed - was the h2 actually being passed as an h1 previously and now is being passed as h3? I imagine my question doesn't make sense but I am looking at the code to the right. Just curious ...
> 
> image
> 
> Anyway, @dartajax approves - great change.

you're spot on. here's the changeset that "downgraded" the `<h1>` to a `<h2>`, not that long ago. https://github.com/ilios/common/commit/dce01d2457f1e6e46c45f99e4ae10004a383c679

PR #3465, opened 2023-07-24
Htmleditor setfocus

stopfstedt:
refs https://github.com/ilios/ilios/issues/4360


jrjohnson (approved):
Nice! It set's focus at the end of text, but since we're using the API of the editor I'm going to assume it does that on purpose.

PR #3486, opened 2023-08-04
Secure our fontawesome token

jrjohnson:
This is annoying, but it appears that maybe someone was using our token to pull packages. We'll need to secure it and provide it to anyone who wants to install Ilios locally. I can find no other way to make this work at this point.


jrjohnson:
Closing in favor of a branch locally as forks can't access secrets.

PR #3489, opened 2023-08-07
replaces all elements that have the button role with actual buttons.

stopfstedt:
then, it reworks markup and styles of expandable/collapsable components, and adds aria-expanded attributes, since the components affected are mostly the same, and these addresses related a11y concepts.

this is all prep work to adding focus state to the app.

refs https://github.com/ilios/ilios/issues/4782


jrjohnson (approved):
This is my favorite thing, excellent!

jrjohnson:
@dartajax you may not be able to merge this (we've got some test issues), but please review!

PR #3493, opened 2023-08-09
bring back scroll-into-view for calendars.

stopfstedt:
this reverts commit 8c81eba3b3aaf2eea50160ce8d461f340548cb4d

refs https://github.com/ilios/ilios/issues/4867 and https://github.com/ilios/common/pull/3460

side-note: i tried the native [`element.scrollIntoView()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) method as suggested by @jrjohnson, in various configurations. these work just as well as the [`scroll-into-view`](https://www.npmjs.com/package/scroll-into-view) package that we're using throughout the frontend. however, it doesn't solve the general problem of the viewport shifting downwards, so this ends up being a straight-up reversal of https://github.com/ilios/common/pull/3460

### DIFF
--- a/addon/components/daily-calendar.hbs
+++ b/addon/components/daily-calendar.hbs
@@ -20,6 +20,8 @@
   <div
     class="day"
     tabindex="0"
+    {{did-insert (perform this.scrollView) this.earliestHour}}
+    {{did-update (perform this.scrollView) this.earliestHour}}
   >
     <div class="hours">
       {{#each this.hours as |hour|}}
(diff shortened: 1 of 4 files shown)


jrjohnson (approved):

dartajax:
Is there no way we can just leave this as it currently is on production?

stopfstedt:
> Is there no way we can just leave this as it currently is on production?

unfortunately not. we need to bring the "scroll first event of the week/day" back, this functionality takes priority over the viewport-shifting issue that this change will reintroduce.

PR #3497, opened 2023-08-14
Adjust Percy Tests

jrjohnson:
I locked a few dates to lower the number of false positive failure rates we have as well as removing some tests that were very close to or identical with other tests.


stopfstedt (approved):
ok cool

PR #3505, opened 2023-08-18
sets the publication check target as scroll target.

stopfstedt:
previously, the scroll-to effect was targeting the container element, which includes the read-only session details above the publication check results table. that wasn't quite cutting it - we want to scroll the results table into view.

fixes https://github.com/ilios/ilios/issues/4922

### DIFF
--- a/addon/components/session-publicationcheck.hbs
+++ b/addon/components/session-publicationcheck.hbs
@@ -1,6 +1,5 @@
 <div
   class="session-publicationcheck"
-  {{scroll-into-view}}
   data-test-session-publicationcheck
 >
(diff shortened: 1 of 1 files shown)


jrjohnson (approved):

dartajax:
I'd be glad to merge this but two browserstack failures (may not prevent merging) and one floating dependency issue (may be preventing the merging) is ensuring that currently I can't merge this - should I do "Enable auto-merge" or re-start the tests?

jrjohnson:
It's the floating one 😢, I went ahead and merged assuming your comment as approval.

PR #3511, opened 2023-08-25
Use browser native fetch

jrjohnson:
We not longer need to get this from a polyfill, all of our targeted browsers have fetch included.


jrjohnson:
nope

PR #3513, opened 2023-08-25
Update our dependency markers

jrjohnson:
NPM wants to do this and align our dependencies with our peer deps, seems reasonable.


jrjohnson:
Nope

PR #3514, opened 2023-08-25
Replace Our Token Library

jrjohnson:
I'd love to keep using `ember-simple-auth-token`, but we're having trouble keeping it up to
date. Doing this work ourselves is reasonable as most of the code can be
removed since we don't support many of the options supported by the
addon.

Turns out we were getting `ember-fetch` from this library so I had to do a little magic where we were using fetch as well.

Then cleanup on dependencies elsewhere because `npm` picked today to be weird for some reason, or else the removal of this library unlocked some kind of deep other things.

This is ready for review / testing / discussion, but not for merge.

Pre-merge test steps:
- [ ] test this on netlify
- [ ] bring this branch into frontend and test there
- [ ] do a pre-release of the frontend PR and test it on stage (with CAS)


jrjohnson:
Testing is done as much as feasible. We'll do more after getting this merged into frontend before releasing. Ready for code review.

stopfstedt (approved):
code looks good to me.

PR #3518, opened 2023-08-28
scrolls taxonomy manager into view when edit-mode is activated.

stopfstedt:
this affects all the places where we're using the taxonomy manager - sessions, courses, and program-years.
fixes https://github.com/ilios/ilios/issues/3839

image


jrjohnson (approved):

dartajax (approved):
works for me

PR #3530, opened 2023-09-05
Run percy tests against a production build

jrjohnson:
Without this we miss things that are different in production, like the way our CSS is complied. Which is dumb, because that's exactly what we're trying to catch.


jrjohnson:
This isn't going to work. Bummer.

PR #3545, opened 2023-09-13
Serialize false in school config

jrjohnson:
Before sending this value to the API it needs to be a string, however ember-data takes the boolean false and converts it into a null so we need to do a bit of special handling in this case.

Refs ilios/ilios#4953


stopfstedt (approved):
click-tested against `master` and the JSON-API PR branch in backend.
worked in both cases.
LGTM.

PR #3555, opened 2023-09-20
Replace deprecation workflow

jrjohnson:
This addon has been amazing, but I've failed in my attempt to modernize
it and we can duplicate all of it's functionality with our own
code which unlocks our embroider upgrade and seems like a good
tradeoff.

I've also decided to throw on new deprecations. This
will require making updates to dependabot PRs when a new deprecation is
added, but it will also prevent us from stacking up a ton of
deprecations and having to track them down.


stopfstedt (approved):
clean and simple. i accept this tuned down replacement.

PR #3561, opened 2023-09-29
Remove sessionLinkingAdminUi flag

jrjohnson:
We've enable this in production (quite a while ago actually) so it can go away here.


stopfstedt (approved):
i accept this.

PR #3581, opened 2023-10-10
Revert pinned node version

jrjohnson:
With the release of node 18.18.1 we can drop this pinned version as the new release has the fix we were waiting for.

Not sure when Github Actions will get this new release, but opening the PR so we can find out.


jrjohnson:
New runner imagine is released. I'm watching the rollout, when it gets closer to 100% I'll mark this as ready.

stopfstedt (approved):

PR #3602, opened 2023-10-30
Drop randomUUID polyfill for ember-data

jrjohnson:
We don't need this anymore as the randomUUID method was added to Safari starting in 15.4. We currently support 15.x, but there is almost no glbal usage of 15.4, everyone is on 15.6 at least (and probably on to 16 or 17).

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,13 +32,6 @@ module.exports = {
         ],
       },
     },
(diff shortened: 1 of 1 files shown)


jrjohnson:
never mind we still need it in iOS.

PR #3605, opened 2023-11-01
Load UUID polyfil ourselves

jrjohnson:
Pulling this from ember-data is complicating our build, let's load it ourselves.


stopfstedt (approved):
i accept this.

PR #3642, opened 2024-01-04
Load icons from sprite file

jrjohnson:
Instead of using the fontawesome javascript process we can load our icons out of the sprite file and reference them in components. This requires a larger initial download, however the browser will cache that data. I think we can play some
pre-loading games like we do for fonts to push the sprite files along with the initial page request to improve this even more.

This allows us to drop all of the fontawesome addon and move away from our team needing to work to maintain that addon ourselves and should unblock our move to pnpm.


stopfstedt:
the original FaIcon component could take a  `@transform` attribute, whereas the drop-in replacement doesnt.

this leads to some (minor) changes that are not intentional. 

image

image

the affected components are:

```
stefan@nichtsnutz: ~/dev/projects/common on plain-icons
$ grep -r '@transform' addon
addon/components/truncate-text.hbs:      <FaIcon @icon="ellipsis" @transform="down-6"/>
addon/components/editable-field.hbs:                <FaIcon @icon="ellipsis" @transform="down-6" />
```

stopfstedt (changes requested):
see other code comment regarding `@transform` argument. otherwise, this looks good to me.

stopfstedt (approved):

PR #3643, opened 2024-01-05
Consolidate font size information

jrjohnson:
Having this in the mixin meant it was included hundreds of times. CSS is fast so this wasn't a huge deal performance wise, but it makes looking at the dev panel annoying and made me feel like a bad developer. Fixed!


stopfstedt (approved):
sadness removal PR _approved_.

PR #3647, opened 2024-01-05
Fix spacing in lm list notes icon

jrjohnson:
This spacing got dropped at some point in the distant past when fa-icon stopped being applied to all icons.


stopfstedt (approved):
nice find.

PR #3654, opened 2024-01-20
Ember CLI code coverage

stopfstedt:
check it out

https://github.com/ember-cli-code-coverage/ember-cli-code-coverage

to run:

`pnpm run coverage`

the output will be generated in a `coverage` directory in the project's root.

sample output

image


jrjohnson (changes requested):
Looks like it's already in `.gitignore`, but can you add `/coverage` to `.npmignore` as well so we don't accidentally ship it.

jrjohnson (approved):

PR #3663, opened 2024-01-24
Convert to monorepo

jrjohnson:
Moved all the tests into the test-app

- I had to remove the code coverage, hopefully temporarily.
- Moved the linting up to the top level. This isn't the way it looks in the sample ember version, but it's the only way I could get everything working the way I expected.
- Removed dotenv, use real ENV variables to configure


jrjohnson:
I'll need to reconfigure netlify after this is merged, so no click testing unfortunately.

stopfstedt (approved):
LGTM

PR #3670, opened 2024-01-26
Fix tagging workflow

jrjohnson:
The pnpm version command doesn't exist. It forwards to npm for some bizare reason. Since we're using pnpm workspaces and not the npm kind this doesn't work for us. Instead we can use this nice tiny npm package `update-version` to accomplish the same thing and modify the package.json the way we want.


stopfstedt (approved):
`changesets` might be worth getting into down the road, but as a quick and easy solution this is great. 

LGTM.

PR #3673, opened 2024-01-26
Manually extract and build the version information

jrjohnson:
The update-version library doesn't do any incrementing on our behalf, it just sets the values we send it. So we have to combine some other information to set this to the right version.

First we extract the data from the current package.json, then we increment it using the semver library, then we pull that incremented version string apart, and lastly we set the values of those individual parts.


jrjohnson:
Going to try this with `versionup` instead.

PR #3680, opened 2024-02-01
Hide back link for frontend

jrjohnson:
We're changing the name of the frontend app from ilios to frontend as part of migrating to a monorepo setup. For now, hide the back link from both application names.


stopfstedt (approved):
SUPERAPPROVED

PR #3682, opened 2024-02-01
Remove pre-commit

jrjohnson:
This isn't working with pnpm as it builds node_modules in a different way. We may replace it. Or just remember to run lint before each commit.


stopfstedt (approved):
ah bummer.

